PR #6, opened 2016-09-10
Bringing to parity with Images

timholy:
This finishes off some API loose ends---including going back to being able to specify padding using strings---and incorporates `extrema_filter`. I copied over @codles implementation and then did a fairly big rewrite. The new version is about twice as fast as the original, not quite as much of a speedup as I was hoping for but nice nonetheless.

This also adds docstrings and docs, so this is basically "done." It will fail CI without supporting PRs merged and packages registered/tagged, so that's the next task.


timholy:
@codles, for the commit I credited to you it seems to have a different email than the one currently associated with your account, even though I copied it from your commit to Images. If you'd like me to change this, please let me know.

PR #10, opened 2016-09-25
Implement changes needed to serve as a foundation for Images

timholy:
This also fixes one fairly serious bug in the algorithm for "filter cascades" (used to handle separable kernels).

One question for the community: in this PR, `extrema_filter` has been turned into a (deprecated) case of a new function, `rankfilter` (it's currently the only thing `rankfilter` does). I'm inclined to rework this yet again, however, calling it something like `blockfilter(f, img, window)`, where the default fallback is to pass blocks of the image to `f`. Does this seem reasonable? Since we can dispatch on function types, there would be no barrier to implementing special algorithms for things like

- `blockfilter(extrema, img, window)` (the equivalent of the old `extrema_filter`), `blockfilter(min, img, window)`, `blockfilter(max, img, window)`
- `blockfilter(median, img, window)`

and others. But the advantage of the fallback is that you could pass any function that takes an array input and returns a scalar value, which is much more general than implied by `rankfilter`. For example, `blockfilter(std, img, window)` to compute the local standard deviation. And it would let us immediately support (slow) versions of long-missing functions like `medfilt`.

Seem reasonable? Any better ideas?


timholy:
I decided to go with `mapwindow` (map a function `f` over sliding windows of the image) instead of `blockfilter`.

Issue #9, opened 2016-09-20
I'm greedy, want more speed!


GunnarFarneback:
There's lots of nice features in this package but benchmarking my use cases didn't turn out as well as I had hoped. To keep things simple, let's limit this discussion to single threaded FIR filtering of a 1000x1000 Float32 image with a non-separable 7x7 Float32 kernel, border replication.

My baseline is a simplified version of an old imfilter implementation I have posted before; you might recognize it. More specifically run this:
```julia
using ImageFiltering
using BenchmarkTools

function my_padarray{T}(img::Matrix{T}, pad1, pad2)
    I1 = clamp((1 - pad1):(size(img, 1) + pad1), 1, size(img, 1))
    I2 = clamp((1 - pad2):(size(img, 2) + pad2), 1, size(img, 2))
    return img[I1, I2]
end

function my_imfilter{T}(img::Matrix{T}, kernel::Matrix{T})
    m, n = size(kernel)
    A = my_padarray(img, m ÷ 2, n ÷ 2)
    offsets = ((0:m-1) - (m - 1) ÷ 2) .+ ((0:n-1) - (n - 1) ÷ 2)' * size(A, 1)
    I = kernel .!= 0
    kernel_values = kernel[I]
    kernel_offsets = offsets[I]
    start_index = m ÷ 2 + (n ÷ 2) * size(A, 1)
    return low_imfilter(A[:], kernel_values, kernel_offsets, size(A, 1),
                        size(A, 2), size(img, 1), size(img, 2), start_index)
end

function low_imfilter{T}(A::Vector{T}, coefficients::Vector{T},
                         offsets::Vector{Int}, in_height::Int, in_width::Int,
                         out_height::Int, out_width::Int, start_index::Int)
    B = zeros(T, out_height, out_width)
    num_coefficients = length(coefficients)
    for n = 1:out_width
        for m = 1:out_height
            index = m + (n - 1) * in_height + start_index
            sum = zero(T)
            for k = 1:num_coefficients
                @inbounds sum += A[index + offsets[k]] * coefficients[k]
            end
            @inbounds B[m, n] = sum
        end
    end
    return B
end

img = rand(Float32, 1000, 1000)
raw_kernel = rand(Float32, 7, 7)
kernel = centered(raw_kernel)
out1 = imfilter(img, kernel, Pad(:replicate), Algorithm.FIR())
out2 = my_imfilter(img, raw_kernel)
[...]

PR #12, opened 2016-10-19
Fix tile sizes and temporary allocation issues in FIR algorithms

timholy:
This fixes the bug noted in https://github.com/JuliaImages/ImageFiltering.jl/issues/9#issuecomment-254013441. The inaccuracy noted in the same issue (https://github.com/JuliaImages/ImageFiltering.jl/issues/9#issuecomment-249136104) was presumably fixed in f66e387c0460c18c3010a328e6eab10bba0aaf87, but that fix had a bug that led to the "inadequate" error.

This also introduces `FIRTiled` so it's possible to control explicitly whether the tiled algorithm is used. If nothing else, this is very useful for testing to make sure that all code-paths get exercised.

CC @GunnarFarneback, @tlnagy.

### DIFF
--- a/src/ImageFiltering.jl
+++ b/src/ImageFiltering.jl
@@ -23,10 +23,11 @@ module Algorithm
     abstract Alg
     "Filter using the Fast Fourier Transform" immutable FFT <: Alg end
     "Filter using a direct algorithm" immutable FIR <: Alg end
+    "Cache-efficient filtering using tiles" immutable FIRTiled <: Alg end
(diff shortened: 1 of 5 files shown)


tlnagy:
I can comfirm that this branch fixes the bug that I ran into, but the performance differences that Gunnar noted are still there. Thanks for working on this!

timholy:
Yeah. Locally I have a branch that gives the best performance of all the options, but not by a lot and so I'm not happy yet. More digging required.

PR #14, opened 2017-01-26
Implement a generic `mapwindow` (e.g., medfilt)

timholy:
We've long had several "holes" in Images, of which one is not having a median filter. This PR implements a generic `mapwindow(f, img, window)` for performing linear or nonlinear operations on rectangular "moving windows" of the image. This is an `O(N^d)` algorithm (where `d` is dimensionality and `N` the window size), so it's not efficient for big windows, but it behaves pretty well for small windows.

For example, let's try `median!`:
```julia
julia> img = rand(1000,1000);

julia> @time mapwindow(median!, img, (3,3));
  0.179863 seconds (144.01 k allocations: 13.309 MB)
```
which is within 30% of Matlab on my machine (and I suspect Matlab is using multithreading, though I didn't check).

The nice thing about this implementation is that in principle, you can pass *any* function to `mapwindow`: e.g., `maximum`, `quantile`, `std`. Functions like `quantile`, which only support `AbstractVector`s, need a little extra care (as described in the docstring).

This doesn't preclude a special-purpose implementation for something like `median!` that gets better than `O(N^d)` behavior (indeed, `extrema` already has an efficient specialization), but I felt the most important thing was to get a flexible, generic algorithm in place.

### DIFF
--- a/src/border.jl
+++ b/src/border.jl
@@ -289,9 +289,6 @@ padarray(img::AbstractArray, f::Fill) = padarray(eltype(img), img, f)
     padindex(border::Pad, lo::Integer, inds::AbstractUnitRange, hi::Integer)
 
 Generate an index-vector to be used for padding. `inds` specifies the image indices along a particular axis; `lo` and `hi` are the amount to pad on the lower and upper, respectively, sides of this axis. `border` specifying the style of padding.
-
-You can specialize this for custom `Pad{:name}` types to generate
-arbitrary (cartesian) padding types.
 """
 function padindex(border::Pad, lo::Integer, inds::AbstractUnitRange, hi::Integer)
     if border.style == :replicate
@@ -308,6 +305,11 @@ function padindex(border::Pad, lo::Integer, inds::AbstractUnitRange, hi::Integer
     else
         error("border style $(border.style) unrecognized")
     end
+end
+function padindex(border::Pad, inner::AbstractUnitRange, outer::AbstractUnitRange)
+    lo = max(0, first(inner)-first(outer))
+    hi = max(0, last(outer)-last(inner))
+    padindex(border, lo, inner, hi)
 end
 
 function inner(lo::Integer, inds::AbstractUnitRange, hi::Integer)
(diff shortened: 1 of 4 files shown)


Richard-Ott:
Is there a way to omit nan in the window? I'm thinking about migrating a maltab code, but need to know first...
Thanks

johnnychen94:
```julia
no_nan(x::NaN) = 0
no_nan(x) = x
mapwindow(x->no_nan(f(x)), img)
```

See also [`ImageCore.clamp01nan`](https://juliaimages.org/latest/function_reference/#ImageCore.clamp01nan)

Richard-Ott:
Doesnt this just change the NaN values to zero? I need them to be omitted in the median calculation.

johnnychen94:
It could start from something like `_median(patch) = median(filter(x->!isnan(x)), patch)` and try to optimize the memory allocation.

Richard-Ott:
Thanks for your input. 
But if I understand correctly, there's no simple solution here and I would need to rebuild a small median filter to be able to omit NaNs. The point being that I'm not used to Julia programming at all and only considered it for speed reasons, but I might search elsewhere if there's no simple solution...

timholy:
With just a move of one parenthesis,
```julia
_median(patch) = median(filter(x->!isnan(x), patch))
```
solves your request:
```julia
julia> using ImageFiltering

julia> img = float.(reshape(1:25, 5, 5))
5×5 Matrix{Float64}:
 1.0   6.0  11.0  16.0  21.0
 2.0   7.0  12.0  17.0  22.0
 3.0   8.0  13.0  18.0  23.0
 4.0   9.0  14.0  19.0  24.0
 5.0  10.0  15.0  20.0  25.0

julia> img[2,2] = NaN
NaN

julia> imgm = mapwindow(_median, img, (3,3))
5×5 Matrix{Float64}:
 1.5   6.0  11.5  16.0  21.0
 2.5   7.0  12.5  17.0  21.0
 3.5   8.5  13.5  18.0  22.0
 5.0   9.0  14.0  19.0  23.0
 5.0  10.0  15.0  20.0  24.0
```

PR #19, opened 2017-01-31
explicitly build docs on 0.5

### LINKED ISSUE #18: 404 page
this  link https://juliaimages.github.io/ImageFiltering.jl/stable/@ref
does not exist
ps. mentioned here 
https://juliaimages.github.io/ImageFiltering.jl/stable/

tlnagy:
Documenter.jl defaults to the nightly linux build (see https://juliadocs.github.io/Documenter.jl/latest/lib/public.html#Documenter.deploydocs), which is more likely to break than the stable linux build. Fixes #18


timholy:
Good idea, thanks!

I wonder if I should delete the docs here, though, since they are also at [JuliaImages](http://juliaimages.github.io/latest/imagefiltering.html#ImageFiltering.jl-1) (including the function reference). Thoughts?

Issue #17, opened 2017-01-30
does not handle integer ...why???


postoroniy:
I have data in integer and integer kernel.
after run imfilter on such data got this
ERROR: InexactError()
 in unsafe_setindex! at /home/slava/.julia/v0.5/OffsetArrays/src/OffsetArrays.jl:153 [inlined]
 in __imfilter_inbounds!(::ComputationalResources.CPU1{ImageFiltering.Algorithm.FIRTiled{2}}, ::Array{Int16,2}, ::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, ::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, ::ImageFiltering.NoPad{ImageFiltering.Pad{0}}, ::CartesianRange{CartesianIndex{2}}, ::Float64) at /home/slava/.julia/v0.5/ImageFiltering/src/imfilter.jl:532
 in _imfilter_inbounds!(::ComputationalResources.CPU1{ImageFiltering.Algorithm.FIRTiled{2}}, ::Array{Int16,2}, ::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, ::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, ::ImageFiltering.NoPad{ImageFiltering.Pad{0}}, ::Tuple{UnitRange{Int64},UnitRange{Int64}}) at /home/slava/.julia/v0.5/ImageFiltering/src/imfilter.jl:499
 in imfilter!(::ComputationalResources.CPU1{ImageFiltering.Algorithm.FIRTiled{2}}, ::Array{Int16,2}, ::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, ::OffsetArrays.OffsetArray{Float64,2,Array{Float64,2}}, ::ImageFiltering.NoPad{ImageFiltering.Pad{0}}, ::Tuple{UnitRange{Int64},UnitRange{Int64}}) at /home/slava/.julia/v0.5/ImageFiltering/src/imfilter.jl:483
[...]

PR #22, opened 2017-02-10
Detect overflow, but don't widen by default

timholy:
Ref #17. This is basically a redo of #20.

### DIFF
--- a/src/imfilter.jl
+++ b/src/imfilter.jl
@@ -140,7 +140,25 @@ function imfilter!(out::AbstractArray, img::AbstractArray, kernel::ProcessedKern
 end
 
 function imfilter!(out::AbstractArray, img::AbstractArray, kernel::ProcessedKernel, border::BorderSpecAny, alg::Alg)
-    imfilter!(default_resource(alg_defaults(alg, out, kernel)), out, img, kernel, border)
+    local ret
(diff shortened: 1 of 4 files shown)


timholy:
Here's the new behavior:
```julia
julia> using ImageFiltering
INFO: Recompiling stale cache file /home/tim/.julia/lib/v0.5/ImageFiltering.ji for module ImageFiltering.

julia> v = fill(0xff, 10)
10-element Array{UInt8,1}:
 0xff
 0xff
 0xff
 0xff
 0xff
 0xff
 0xff
 0xff
 0xff
 0xff

julia> kern = centered(fill(0xff, 3))
OffsetArrays.OffsetArray{UInt8,1,Array{UInt8,1}} with indices -1:1:
 0xff
 0xff
 0xff

julia> imfilter(v, kern)
WARNING: Likely overflow or conversion error detected. Consider specifying the output type, e.g., `imfilter(UInt32, img, kernel, ...)`
ERROR: InexactError()
 in unsafe_setindex! at /home/tim/.julia/v0.5/OffsetArrays/src/OffsetArrays.jl:158 [inlined]
# rest of backtrace...
```
With a helpful error message, any pain for the users should be very minimal.

PR #28, opened 2017-03-29
Fix remaining errors on Julia 0.6

timholy:
The fact that this is possible is thanks to the new StaticArrays.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,12 +9,12 @@ julia:
 notifications:
   email: false
 # uncomment the following lines to override the default test script
-# script:
-#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-#  - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageFiltering"); Pkg.test("ImageFiltering"; coverage=true)'
+script:
+ - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+ - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageFiltering"); Pkg.test("ImageFiltering"; coverage=VERSION >= v"0.6.0-alpha")'
 after_success:
   # push coverage results to Codecov
-  - julia -e 'cd(Pkg.dir("ImageFiltering")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
+  - julia -e 'if VERSION >= v"0.6.0-alpha" cd(Pkg.dir("ImageFiltering")); Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder()); end'
   # update the documentation
   - julia -e 'Pkg.add("Documenter")'
   - julia -e 'cd(Pkg.dir("ImageFiltering")); ENV["DOCUMENTER_DEBUG"] = "true"; include(joinpath("docs", "make.jl"))'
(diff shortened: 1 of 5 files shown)


timholy:
So unfair. https://github.com/JuliaMath/FixedPointNumbers.jl/pull/81 cropped up in the 2 day gap since I last built julia master.

The more surprising error is on 0.5. I have checked that I have exactly the same version of those packages, and that I have exactly the same julia commit. For me that test throws a `TypeError`, but it doesn't error on Travis. Any ideas?

Evizero:
The tests of this branch pass for me on 0.5

```
julia> versioninfo()
Julia Version 0.5.0
Commit 3c9d753* (2016-09-19 18:14 UTC)
Platform Info:
  System: Darwin (x86_64-apple-darwin15.6.0)
  CPU: Intel(R) Core(TM) M-5Y51 CPU @ 1.10GHz
  WORD_SIZE: 64
  BLAS: libopenblas (DYNAMIC_ARCH NO_AFFINITY Haswell)
  LAPACK: libopenblas
  LIBM: libopenlibm
  LLVM: libLLVM-3.7.1 (ORCJIT, broadwell)
```

I am on the tagged version of most packages

Evizero:
Well one guess I could make is: Are you on 0.5.0 or 0.5.1 locally? If you are on 0.5.0 like I am then maybe that is the culprit

timholy:
Exact same commit that Travis is using. (Both 0.5.1.)

timholy:
Turns out to be a difference that depends on whether inlining is on or not. By default we turn it off when measuring coverage; however, this doesn't happen for 0.6, so there's a happy workaround.

Ref https://github.com/JuliaLang/julia/issues/21238.

.travis.yml
```diff
@@ -9,12 +9,12 @@ julia:
 notifications:
   email: false
 # uncomment the following lines to override the default test script
-# script:
-#  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-#  - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageFiltering"); Pkg.test("ImageFiltering"; coverage=true)'
+script:
+ - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+ - julia -e 'Pkg.clone(pwd()); Pkg.build("ImageFiltering"); Pkg.test("ImageFiltering"; coverage=VERSION >= v"0.6.0-alpha")'
```
tkelman: `-pre.alpha`, since it had to sort after `-dev`

timholy: Hah, funny. Still, it works so I think I'll keep it this way.

test/border.jl
```diff
@@ -141,7 +141,8 @@ using Base.Test
     @testset "Pad" begin
         @test Pad(:replicate,[1,2], [5,3]) == Pad(:replicate,(1,2), (5,3))
         @test @inferred(Pad{2}(:replicate, [1,2], [5,3])) == Pad(:replicate,(1,2), (5,3))
-        @test_throws ErrorException Pad{3}(:replicate, [1,2], [5,3])
+        paderr = VERSION < v"0.6.0-alpha" ? TypeError : MethodError
```
tkelman: the travis one doesn't matter much, but this one does - what changed the behavior here?

timholy: The type-system rewrite?

PR #32, opened 2017-05-04
add check for kernel ndims [WIP]

SimonDanisch:
I tried to add checks to assert that kernel and image dimensionality matches.... Sadly:
image
Could someone knowing more about the types of kernels there are help me implement `kernel_ndims` ?

Best,
Simon

### DIFF
--- a/src/imfilter.jl
+++ b/src/imfilter.jl
@@ -1,4 +1,15 @@
 # see below for imfilter docstring
+kernel_ndims(kernel::Tuple) = ndims(first(kernel))
+kernel_ndims(kernel::AbstractArray) = ndims(kernel)
+function assert_ndims(array, kernel)
+    n1, n2 = ndims(array), kernel_ndims(kernel)
+    if n1 != n2
+        error(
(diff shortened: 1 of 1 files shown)


timholy:
Currently ImageFiltering treats missing trailing dimensions as singletons. This allows the following:
```julia
using ImageFiltering, TestImages
img = testimage("lighthouse")
imgf = imfilter(img, (k, k'))
k = [1,1,1]/3
imgf = imfilter(img, (centered(k), centered(k')))
```
If we did that dimension check we'd have to change this to
```julia
imgf = imfilter(img, (centered(reshape(k, 3, 1)), centered(k')))
```
That's not terrible, but can you tell me a little more about the kind of error you're trying to catch?

timholy:
In case it's not clear, the tuple allows you to exploit the computational advantages of separable filters:
- https://en.wikipedia.org/wiki/Separable_filter
- http://blogs.mathworks.com/steve/2006/10/04/separable-convolution/

PR #38, opened 2017-08-19
Fix errors on 32bit platforms

timholy:
I recently got a [tip](https://github.com/JuliaLang/julia/pull/23330) on how to compile a 32-bit version of julia on a 64-bit machine. It's long bothered me that the tests here fail on 32-bit, so I took the opportunity to figure out why they are failing and fix the problem.


timholy:
Yay, worked. For nightly we'll have to wait for fixes in Base or StaticArrays.

PR #39, opened 2017-09-14
make sure mapwindow() function reference is generated

rdeits:
The function reference for `mapwindow()` is missing, which makes the link [here](https://juliaimages.github.io/ImageFiltering.jl/stable/index.html#Arbitrary-operations-over-sliding-windows-1) broken. I think we just need to tell Documenter to look inside the MapWindow module.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -1,6 +1,6 @@
 using Documenter, ImageFiltering
 
-makedocs(modules  = [ImageFiltering, Kernel, KernelFactors],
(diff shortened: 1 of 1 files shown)


timholy:
Thanks! Looks like it was generated [here](http://juliaimages.github.io/latest/function_reference.html#ImageFiltering.MapWindow.mapwindow), but if this package is going to have its own docs, they'd better be right!

PR #40, opened 2017-09-19
Circumvent a deprecation warning

timholy:
Finally tracked down the source of the longstanding
```julia
WARNING: deprecated syntax "_ as an rvalue around docs/Docs.jl:522".
```


Evizero:
YES

timholy:
It's interesting that this only happens for functions with docstrings, but I didn't chase it down far enough to understand why. Simple to circumvent, however.

PR #41, opened 2017-09-23
WIP: fix on Julia 0.7

timholy:
Over the last week or so I've finally gotten to the point where the dependent packages seem to work on 0.7, so it's time to tackle the big one.

This breaks 0.6, but the point of pushing this now is that locally it reveals a segfault in Julia master. Seems worth having a record of that.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -7,6 +7,7 @@ OffsetArrays
 CatIndices
 StaticArrays 0.0.5
 MappedArrays
+FFTW 0.0.3
 FFTViews
 ComputationalResources
 DataStructures 0.4.6
--- a/src/ImageFiltering.jl
+++ b/src/ImageFiltering.jl
@@ -2,12 +2,22 @@ __precompile__()
 
 module ImageFiltering
 
+importall FFTW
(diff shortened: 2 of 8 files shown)


Evizero:
You are probably aware of this, but just to be safe: I think `ImageCore` (which this package depends on) still fails on 0.7 according to travis.

timholy:
I wasn't until your PR. Will look into it now.

RalphAS:
I think this can be rebased on master, making the package almost ready for 0.7. It still needs the following:

1. `src/imfilter.jl`, refers to the obsolete `ImageCore.squeeze1`

2. `test/runtests.jl` needs an `import StaticArrays`

3. There are additional problems with indexing into `OffsetArrays` a/o using `AbstractRange`. (Can't do `length` of a `Slice`, hence can't `vcat` two of them in `padarray/padindex`; I don't understand the context well enough to propose a fix.)

But adding (2) and stubbing out (1) above makes this package seem usable (on Julia 0.7.0-beta.123, and with your v0.7-rewrite branch of ImageCore).  There are only a few failures/errors when running a package test in this configuration, most related to the indexing business.

timholy:
Superseded by #65

Issue #44, opened 2017-11-08
Add stride argument for mapwindow


jw3126:
Sometimes, I need to apply a function to a window, which slides over an image. Just like what `mapwindow` does, except that I want my window to move by bigger steps. Would it be an option to add an additional argument to `mapwindow` that allows to select how far the window moves in each dimension? For example in the case of `Matrix` it behaves as follows:
```julia 
mapwindow(f, args...; stride=(3,2)) == mapwindow(f, args..., windowsize)[1:3:end,1:2:end]
```
except that the unnecessary computations are omitted. If there is interest in this I can implement it.

timholy:
I think it's a great idea. The one extra thing that I think needs consideration is how to specify an offset, to achieve the same thing as
```julia
mapwindow(f, args...)[2:3:end,2:2:end]
```
or maybe even
```julia
mapwindow(f, args...)[101:3:128, 40:3,77]
```
to focus on an ROI.
 
We could use both `stride` and `offset` keywords, but I wonder if a better approach would be founded on passing the ranges themselves so that you can specify an extra endpoint.

jw3126:
Yeah full control over the ranges sounds good. And lots of narrow keyword arguments are a mixed blessing. My only problem is, that I am mostly interested in the stride case  and this is awkward to express as ranges.
But I guess this could be solved with a helper function `strided(indices(img), (2,3))`.

jw3126:
If I want to iterate over a product of unit ranges, I can to `CartesianRange(some_unitranges)`. What is a good way to iterate over a product of non unit ranges?

timholy:
There's `Iterators.product`

jw3126:
Closing  #45.

PR #45, opened 2017-11-09
RFC: strided mapwindow

jw3126:
See #44. Here is a first implementation. The API is a bit cumbersome and needs to be improved. Currently you have to call 
```julia
mapwindow(f,img,window,border,imginds)
```
Any suggestions for a better API? I would also like to add two convenience functions:
* `strided(inds, stride)` converts a stride into something that can be passed to mapwindow:
```julia
mapwindow(f,...,strided(indices(img), (2,3))) = mapwindow(f,...)[1:2:end,1:3:end]
```
* `mappool(f,img,window)` which does image [pooling](https://discourse.julialang.org/t/image-pooling/5280). This is the special case, where the stride is the size of the window.
```julia
mappool(f,img,window,...) = mapwindow(f,img,...,strided(indices(img), map(length,window))
```

### DIFF
--- a/src/mapwindow.jl
+++ b/src/mapwindow.jl
@@ -7,7 +7,7 @@ using Base: Indices, tail
 export mapwindow
 
 """
-    mapwindow(f, img, window, [border="replicate"]) -> imgf
+    mapwindow(f, img, window, [border="replicate"], [imginds=indices(img)]) -> imgf
 
 Apply `f` to sliding windows of `img`, with window size or indices
 specified by `window`. For example, `mapwindow(median!, img, window)`
@@ -29,6 +29,14 @@ AbstractUnitRanges, in which case the specified ranges are used for
 `border` specifies how the edges of `img` should be handled; see
 `imfilter` for details.
 
+Finally `imginds` allows to omit unnecessary computations, if you want to do things
+like `mapwindow` on a subimage, or a strided variant of mapwindow.
+It works as follows:
+```julia
+mapwindow(f, img, window, border, (2:5, 1:2:7)) == mapwindow(f,img,window,border)[2:5, 1:2:7]
+```
+Except more efficiently because it omits computation of the unused values.
+
 For functions that can only take `AbstractVector` inputs, you might have to
 first specialize `default_shape`:
 
@@ -41,48 +49,41 @@ and then `mapwindow(f, img, (m,n))` should filter at the 75th quantile.
 
 See also: [`imfilter`](@ref).
 """
-function mapwindow(f, img::AbstractArray, window::Dims, args...; kwargs...)
+function mapwindow(f, img, window, border="replicate", imginds=indices(img); callmode=:copy!)
+    _mapwindow_kernel(replace_function(f),
+              img,
+              resolve_window(window),
+              resolve_border(border),
+              resolve_imginds(imginds),
+              default_shape(f);
+              callmode=callmode)
+end
+
+function resolve_window(window::Dims)
     all(isodd(w) for w in window) || error("entries in window must be odd, got $window")
     halfsize = map(w->w>>1, window)
-    mapwindow(f, img, map(h->-h:h, halfsize), args...; kwargs...)
+    map(h -> -h:h,halfsize)
 end
-function mapwindow(f, img::AbstractVector, window::Integer, args...; kwargs...)
+function resolve_window(window::Integer)
     isodd(window) || error("window must be odd, got $window")
     h = window>>1
-    mapwindow(f, img, (-h:h,), args...; kwargs...)
+    -h:h
 end
+resolve_window(window::AbstractArray) = resolve_window((window...,))
+resolve_window(window::AbstractUnitRange) = (window,)
+resolve_window(window::Indices) = window
 
-mapwindow(f, img::AbstractArray, window::Indices; kwargs...) =
-    mapwindow(f, img, window, "replicate"; kwargs...)
-mapwindow(f, img::AbstractVector, window::AbstractUnitRange; kwargs...) =
-    mapwindow(f, img, (window,); kwargs...)
+resolve_border(border::AbstractString) = borderinstance(border)
+resolve_border(border::BorderSpecAny) = border
 
-function mapwindow(f, img::AbstractArray, window::Indices, border::AbstractString;
-                   kwargs...)
-    mapwindow(f, img, window, borderinstance(border); kwargs...)
-end
-function mapwindow(f, img::AbstractVector, window::AbstractUnitRange, border::AbstractString;
-                   kwargs...)
-    mapwindow(f, img, (window,), border; kwargs...)
-end
+resolve_imginds(r::Range) = (r,)
+resolve_imginds(imginds) = imginds
 
-mapwindow(f, img, window::AbstractArray, args...; kwargs...) = mapwindow(f, img, (window...,), args...; kwargs...)
-
-function mapwindow(f,
-                   img::AbstractArray{T,N},
-                   window::Indices{N},
-                   border::BorderSpecAny,
-                   imginds::NTuple{N, Range}=indices(img);
-                   callmode=:copy!) where {T,N}
-    _mapwindow(replace_function(f), img, window, border, imginds, default_shape(f); callmode=callmode)
-end
 abstract type _IndexTransformer end
 
 struct _AffineTransformer{N} <: _IndexTransformer
     offset::NTuple{N,Int}
     stride::NTuple{N,Int}
-end
-function _AffineTransformer(ranges)
 end
 @inline function Base.getindex(t::_AffineTransformer, inds::CartesianIndex)
     CartesianIndex(t.offset .+ t.stride .* inds.I)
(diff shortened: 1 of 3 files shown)


timholy:
Sorry for the delay here; I've been mulling over the API issues, and I agree it's not entirely straightforward. I keep returning to something like:
```julia
dest = PooledArray{T}(inds)
mapwindow!(f, dest, img, window, border)
```
and possibly a convenience method
```julia
dest = pooledarray(f, img, window, inds)
```
for the automatic computation of the eltype.

But to be honest I'm not convinced that's any better.

test/runtests.jl
```diff
@@ -5,14 +5,14 @@ aif = detect_ambiguities(ImageFiltering, Kernel, KernelFactors, Base)
 asa = detect_ambiguities(StaticArrays, Base)
 @test isempty(setdiff(aif, asa))
 
+include("mapwindow.jl")
```
timholy: Is this deliberate or simply a temporary convenience?

test/mapwindow.jl
```diff
@@ -91,5 +91,18 @@ using ImageFiltering, Base.Test
...
+    function groundtruth(f, A, window::Tuple, border, imginds)
+        mapwindow(f,A,window,border)[imginds...]
+    end
+    for (f,img, window, imginds) ∈ [
+            (mean, randn(10), (1,), (1:2:10,)),
+            (median!, randn(10), (-1:1,), (1:2:8,)),
+            (mean, randn(10), (-1:1,), (1:2:8,)),
+            (mean, randn(10,5), (-1:1,0:0), (1:2:8,1:3)),
+        ]
+        border = ImageFiltering.borderinstance("replicate")
```
timholy: Can we not just pass `"replicate"` directly?

jw3126: Yes we can!

test/runtests.jl
jw3126: Just a convenience, I can restore previous order if this is finished.

jw3126:
None of the `@assert` is inside a hot loop, so I don't think they will have a performance impact. I could turn them into tests, but these are implementations details I think assertions are much more convenient then tests here.

jw3126:
Introducing a new type of array seems a bit heavyweight for this. Is turning `border` into a keyword argument an option?
```julia 
mapwindow(f,img,window,imginds; border=myborder)
```
At least border does not influence the return type AFAICT.
```julia
mapwindow!
```
seems to be useful also independently of this PR.

timholy:
> At least border does not influence the return type AFAICT.

It does for `Fill`, `NoPad`, and `Inner`.

> Introducing a new type of array seems a bit heavyweight for this.

I agree. The only justification is whether it would be useful for "documenting" the results later. For example, if someone hands you an array that a result of some pooling operation, does it help you to know where each "pixel" (result on a chunk) came from? Or do you not care?

And yes, `mapwindow!` is probably independently useful no matter what we decide here.

jw3126:
> does it help you to know where each "pixel" (result on a chunk) came from? Or do you not care?

Depends on the application. Most of the time I just want to detect, whether some property is holds in an image. For this I don't care about the original pixels. But if you search something in an image, this sounds useful. In that case you might also want to chain some `mapwindow!` calls? Not sure. Also if you are interested in the original pixels, would you also like to remember `windowsize`?

Finally one might want to use a strided `mapwindow!` with another array type as `dest`. So we would probably need
```julia
struct PooledArray{T,N,A,I} <: AbstractArray{T,N}
    data::T
    indices::I
end
```
This is even more heavyweight and looks a lot like an `AxisArray`.

Here is a list of all options I can think of:
1. Introduce new array type `PooledArray` that wraps the output and use `mapwindow!`
2. Introduce new array type that wraps the input.
3. `mapwindow(f, img, windowsize, border, indices)`
4. Introduce a new function `getindex_mapwindow(f,img,windowsize, indices, border)`
5. Make `indices` a keyword argument. This would break type stability if we tried to support e.g. `(1:3, 2)` in future.

jw3126:
Another idea is making `mapwindow` lazy.
```julia
struct MapWindow{T,N,F,O,W,P} <: AbstractArray{T,N}
    f::F
    original_array::O
    windowsize::W
    pad::P
    # maybe also buffer?
end

MapWindow(f, img, windowsize)[1:4, 5:6]

mapwindow!(f,dest, img,windowsize, border) = copy!(dest, MapWindow(f,img,windowsize, border))
```

timholy:
I generally love lazy, but here the problem is that it kind of violates one of the implicit "contracts" of `AbstractArray`: that elementwise access is cheap. I added something along these lines to `ImageAxes` in https://github.com/JuliaImages/ImageAxes.jl/pull/19, and ended up not making it a subtype of `AbstractArray`. This of course makes some things more painful, but in general I think it's better to rely on "behavior" than it is on subtypes.

Of your list of 5 options above, I'd like to strike number 2. This is really a statement about what portion of the output you want to keep, so associating it with the input seems wrong. 4 seems somehow unnecessary. I find myself roughly equally attracted by 1, 3, and 5. I can go with the API you have now, if there isn't an obviously better alternative. Since I'm on the fence, I'd be happy to have you choose.

jw3126:
I will go with 
```julia
mapwindow(f,img,window,border,imginds)
```
for now. Maybe over time something better will emerge.

src/mapwindow.jl
```diff
@@ -29,6 +29,13 @@ AbstractUnitRanges, in which case the specified ranges are used for
 `border` specifies how the edges of `img` should be handled; see
 `imfilter` for details.
 
+Finally `imginds` allows to omit unnecessary computations, if you want to do things
+like `mapwindow` on a subimage, or a strided variant of mapwindow.
+It works as follows:
+```julia
+mapwindow(f, img, window, border, (2:5, 1:2:7)) == mapwindow(f,img,window,border)[2:5, 1:2:7]
+```
```
timholy: Perhaps add "except more efficiently because it omits computation of the unused values."

src/mapwindow.jl
```diff
@@ -41,78 +48,176 @@ and then `mapwindow(f, img, (m,n))` should filter at the 75th quantile.
 
 See also: [`imfilter`](@ref).
 """
-function mapwindow(f, img::AbstractArray, window::Dims, args...; kwargs...)
+function mapwindow(f,img,window;kw...)
+    mapwindow(f, img, window, "replicate", kw...)
```
timholy: semicolon after `"replicate"`. Usually stylistic, but here I think it's actually a bug (like it was in https://github.com/JuliaImages/ImageView.jl/pull/130/files).

Also perhaps add spaces between the arguments?

src/mapwindow.jl
```diff
@@ -41,78 +48,176 @@ and then `mapwindow(f, img, (m,n))` should filter at the 75th quantile.
...
+    map(h -> -h:h,halfsize)
 end
-function mapwindow(f, img::AbstractVector, window::Integer, args...; kwargs...)
+function resolve_window(window::Integer)
     isodd(window) || error("window must be odd, got $window")
     h = window>>1
-    mapwindow(f, img, (-h:h,), args...; kwargs...)
+    -h:h
+end
+resolve_window(window::AbstractArray) = resolve_window(img, (window...,))
```
timholy: where does the `img` come from?

src/mapwindow.jl
```diff
@@ -41,78 +48,176 @@ and then `mapwindow(f, img, (m,n))` should filter at the 75th quantile.
...
+function _indices_of_interiour_indices(fullimginds, imginds, kerinds)
+    map(_indices_of_interiour_range, fullimginds, imginds, kerinds)
 end
-function _mapwindow(f,
+
+# replace median by ... outside of _mapwindow_kernel
+function _mapwindow_kernel(f,
                     img::AbstractArray{T,N},
-                    window::Indices{N},
+                    window::NTuple{N,Range},
```
timholy: How about an `AbstractUnitRange` here, since I think we wouldn't properly handle a non-1 step if it were supplied.

jw3126:
Thanks for the feedback. There are still some  corner cases, that I need to test/fix.

jw3126:
If there are no further objections, I will merge on Monday.

test/mapwindow.jl
```diff
@@ -91,5 +91,22 @@ using ImageFiltering, Base.Test
...
+            (mean, randn(10), (-1:1,), (1:2:8,)),
+            (mean, randn(10,5), (-1:1,0:0), (1:2:8,1:3)),
+            (mean, randn(10,5), (-1:1,0:0), (Base.OneTo(2),1:3)),
+        ]
+        border = "replicate"
+        @test groundtruth(f,img,window,border,imginds) == @inferred mapwindow(f,img,window,border,imginds)
+    end
+    @test mapwindow(mean, randn(10), (3,), "replicate", 2:2:7) isa Array
+    @test mapwindow(mean, randn(10), (3,), "replicate", 2:7) isa OffsetArray
+    @test mapwindow(mean, randn(10), (3,)) isa Array
```
timholy: Sorry I didn't notice this earlier, but over time I've gotten more skeptical about testing the output type directly (this hit me with a vengance in https://github.com/JuliaImages/ImageCore.jl/pull/52). Can something equivalent be done by testing the `indices`?

jw3126: Yeah I don't like these tests either. Testing the indices is more sane.

timholy:
Thanks so much, this is fantastic!

PR #47, opened 2017-11-15
add mapwindow benchmark

jw3126:
Useful to judge performance impact of #45. Are there more `mapwindow` cases, that should be benchmarked?


timholy:
Nothing more occurred to me. This is great, many thanks!

PR #49, opened 2017-11-20
add mapwindow!

jw3126:


src/mapwindow.jl
```diff
@@ -372,12 +422,6 @@ end
 # This is slightly faster than a circular buffer
 @inline cyclecache(b, x) = b[1], (Base.tail(b)..., x)
 
-replace_function(f) = f
-replace_function(::typeof(median!)) = function(v)
```
jw3126: What does `replace_function` do? Is it still needed? Tests still pass if I delete it.

src/mapwindow.jl
```diff
@@ -49,16 +50,47 @@ and then `mapwindow(f, img, (m,n))` should filter at the 75th quantile.
...
+                   imginds=default_imginds(img, window, border); callmode=:copy!)
+    if callmode != :copy!
+        error("Only callmode=:copy! is currently supported")
+    end
+    _mapwindow(f,
               img,
               resolve_window(window),
               resolve_border(border),
-              resolve_imginds(imginds),
-              default_shape(f);
```
timholy: `default_shape` is used as needed to reshape `buf` before passing it to `f`. One could require that `f` handle whatever shape you pass to it: if the innate implementation of `f` requires an `AbstractVector` input, one could define `f(A) = f(vec(A))`.

Unfortunately, (at least at one point) this led to a performance degradation, because this approach requires that you call `vec` for every output point. The original implementation of `mapwindow` used `vec` once to create a view of `buf` and then passed the view to `f`.

If this wasn't clear, then I should have added more comments. Sorry about that.

jw3126: Yes, `default_shape` still exists, it is used in the `allocate_buffer` function, instead of `mapwindow_kernel!` directly.

src/mapwindow.jl
```diff
@@ -165,47 +197,64 @@ function _indices_of_interiour_indices(fullimginds, imginds, kerinds)
...
+end
+
+function compute_output_eltype(f, img, window, border, imginds)
+    buf, bufrs = allocate_buffer(f, img, window)
+    make_buffer_values_realistic!(buf, img, window, border, imginds)
+    typeof(f(bufrs))
+end
+
+function make_buffer_values_realistic!(buf, img, window, border::Inner, imginds)
+    buf
```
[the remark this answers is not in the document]
jw3126: Yes. A hacky solution would be to fill the buffer with `one(eltype(img))`. A better solution would be to fill the buffer with data from the image. Doing that is a bit convoluted, since we need to make sure, that we stay in the inner of the image.

jw3126:
Okay, I saw that the replaced `median!` is faster, but wondered why a faster version lives here and not in `Base`. The comment about `keyword` arguments explains it. Can we get rid of this, as soon as `Base` has fast keywords?

src/mapwindow.jl
jw3126: Okay, I saw that the replaced `median!` is faster, but wondered why a faster version lives here and not in `Base`. The comment about `keyword` arguments explains it. Can we get rid of this, as soon as `Base` has fast keywords?

src/mapwindow.jl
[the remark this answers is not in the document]
jw3126: Are you okay with filling the buffer with `one(eltype(img))`?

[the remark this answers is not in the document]
timholy: Change that to `oneunit` and I think you have a deal. I only know of one case where this would fail:
```julia
julia> using FixedPointNumbers

julia> oneunit(Q0f7)
ERROR: InexactError()
Stacktrace:
 [1] convert at /home/tim/.julia/v0.6/FixedPointNumbers/src/fixed.jl:42 [inlined]
 [2] one at ./number.jl:163 [inlined]
 [3] oneunit(::Type{FixedPointNumbers.Fixed{Int8,7}}) at ./number.jl:186

julia> typemax(Q0f7)
0.992Q0f7
```

src/mapwindow.jl
```diff
@@ -84,13 +85,24 @@ If `out` and `img` have overlapping memory regions, behaviour is undefined.
...
+
+function median_fast!(v)
+    # median! calls partialsort! which has keyword arguments. Keyword arguments are slow.
+    # This replaces median! with a more efficient implementation free of keyword arguments.
+    inds = indices(v,1)
+    Base.middle(Base.select!(v, (first(inds)+last(inds))÷2, Base.Order.ForwardOrdering()))
+end
+
+resolve_f(f) = f
+resolve_f(::typeof(median!)) = median_fast!
```
timholy: If we're changing the name maybe add `@deprecate_binding replace_function resolve_f`? This is something that users may be extending in their own code (it was designed that way).

src/mapwindow.jl
```diff
@@ -1,10 +1,11 @@
 module MapWindow
 
 using DataStructures, TiledIteration
-using ..ImageFiltering: BorderSpecAny, Pad, Fill, borderinstance, _interior, padindex, imfilter
+using ..ImageFiltering: BorderSpecAny, Pad, Fill, Inner,
+borderinstance, _interior, padindex, imfilter
```
timholy: indentation here too

src/mapwindow.jl
jw3126: Okay I thought it was an implementation detail. If it is public, I don't think its worth renaming.

timholy: It's not really "public" (because you'd have to `import` or scope it to extend it), but was definitely intended to be exploited by anyone brave enough to dig into the implementation :slightly_smiling_face: .

timholy:
Thanks again!

PR #61, opened 2018-04-30
Patches `padarray` so that it works on `OffsetArray`

zygmuntszpak:
Using the `padarray` function with the `Fill` type on an `OffsetArray` resulted in an error.  This was caused due to an implicit call of `size` on an `OffsetArray` which is an undefined operation. This commit patches the bug JuliaImages/Images.jl/issues/718

### DIFF
--- a/src/border.jl
+++ b/src/border.jl
@@ -902,7 +902,12 @@ function padarray(::Type{T}, img::AbstractArray{S,N}, f::Fill{_,N}) where {T,S,_
     catch
         error("Unable to fill! an array of element type $(eltype(A)) with the value $(f.value). Supply an appropriate value to `Fill`, such as `zero(eltype(A))`.")
     end
-    A[indices(img)...] = img
+
+    # We would have liked to do A[indices(img)...] = img, but this calls size(img)
+    # which is not defined if img is of type OffsetArrays.
+    for I in Compat.CartesianIndices(Compat.axes(img))
+        A[I] = img[I]
+    end
     A
 end
 padarray(img::AbstractArray, f::Fill) = padarray(eltype(img), img, f)
--- a/test/border.jl
+++ b/test/border.jl
@@ -62,6 +62,66 @@ using Base.Test
             ret = @test_throws ArgumentError padarray(A, Pad(Style))
             @test contains(ret.value.msg, "lacks the proper padding")
         end
+        # same as above, but input arrays are of type OffsetArray
+        Ao =  OffsetArray(reshape(1:25, 5, 5),-1,-1)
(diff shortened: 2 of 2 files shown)


src/border.jl
```diff
@@ -902,7 +902,12 @@ function padarray(::Type{T}, img::AbstractArray{S,N}, f::Fill{_,N}) where {T,S,_
     catch
         error("Unable to fill! an array of element type $(eltype(A)) with the value $(f.value). Supply an appropriate value to `Fill`, such as `zero(eltype(A))`.")
     end
-    A[indices(img)...] = img
+
+    # We would have liked to do A[indices(img)...] = img, but this calls size(img)
+    # which is not defined if img is of type OffsetArrays. 
+    for i in collect(Iterators.product(OffsetArrays.axes(img)...))
+             setindex!(A,getindex(img,i...),i...)
+    end
```
Evizero: How about 

```julia
for I in CartesianRange(indices(img))
    A[I] = img[I]
end
```

src/border.jl
```diff
@@ -902,7 +902,12 @@ function padarray(::Type{T}, img::AbstractArray{S,N}, f::Fill{_,N}) where {T,S,_
     catch
         error("Unable to fill! an array of element type $(eltype(A)) with the value $(f.value). Supply an appropriate value to `Fill`, such as `zero(eltype(A))`.")
     end
-    A[indices(img)...] = img
+
+    # We would have liked to do A[indices(img)...] = img, but this calls size(img)
+    # which is not defined if img is of type OffsetArrays. 
+    for i in collect(Iterators.product(OffsetArrays.axes(img)...))
```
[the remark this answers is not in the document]
zygmuntszpak: Thanks for the clarification.

src/border.jl
zygmuntszpak: I used `@which Compat.axes(img)` for both a traditional array and Offset array and it seems to be calling the same functions as `OffsetArrays.axes(img)`. Why do we use `Compat.axes(img)` in this instance and not `OffsetArrays.axes(img)`? I thought the `Compat` package has something to do with deprecations?

zygmuntszpak:
I've changed it to your suggestion, but I'm wondering why we ought to use `Compat.axes` and not `OffsetArrays.axes` in this instance?

Evizero:
Because as far as I can tell that's where the `axes` actually comes from. `OffsetArrays.axes` works because that package also uses that function. As I understand `axes` is the 0.7 version of `indices`, so for backward compatibility `Compat` provides some glue code to also make it work in 0.6. It seems in 0.6 its just a reference to `indices`

at least thats my impression when looking at https://github.com/JuliaArrays/OffsetArrays.jl/commit/ce854a71a9e96eba081d48144b9df66928ebdc91

PR #67, opened 2018-07-21
A few more patches for v0.7

RalphAS:
This set of changes makes more tests work, gets some common filtering calls from other packages working, and obviates the deprecations observed by me so far.  The "tests work" part depends on some pending PRs to TiledIterations.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,9 +1,17 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
-  - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
-  - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
+  - julia_version: 0.7
+  - julia_version: latest
+
+platform:
+  - x86 # 32-bit
+  - x64 # 64-bit
+
+## uncomment the following lines to allow failures on nightly julia
+## (tests will run but not make your overall status red)
+matrix:
+  allow_failures:
+  - julia_version: latest
 
 branches:
   only:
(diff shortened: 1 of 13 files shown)


RalphAS:
Without the `EdgeIterator` patch, the following tests error out because of a bounds error (for 4/5 of the test inputs):
https://github.com/JuliaImages/ImageFiltering.jl/blob/master/test/mapwindow.jl#L113
https://github.com/JuliaImages/ImageFiltering.jl/blob/master/test/mapwindow.jl#L115

Although https://github.com/JuliaArrays/TiledIteration.jl/pull/14 works around the problem, it may just be hiding a mistake in the `mapwindow` revision. (Specifically, computation of output indices in `mapwindow_kernel!`)

Evizero:
Any guess as why it didn't error in 0.6?

RalphAS:
I've had a chance to study the problem again and put the responsibility back in TiledIteration.  See an improved PR: https://github.com/JuliaArrays/TiledIteration.jl/pull/15

timholy:
Sorry for the delay here, my preparation for JuliaCon is eating all my time.

Have you tested since https://github.com/JuliaLang/julia/pull/28401? I wrote that after poking at this a bit, but I never got back to the motivating issue. It might have been just something that came up in my debugging rather than something that was actually causing the error in this package.

RalphAS:
Tests marked broken (real breakage for LoG and failed inference in gradients) apply to v0.7-rc2 which IIUC has the patch you mention.

src/imfilter.jl
```diff
@@ -1547,7 +1547,7 @@ function normalize_separable!(r::AbstractResource, A, kernels::NTuple{N,TriggsSd
     function imfilter_inplace!(r, a, kern, border)
         imfilter!(r, a, a, (kern,), border)
     end
-    filtdims = ntuple(d->imfilter_inplace!(r, fill(1, inds[d]), kernels[d], border), Val(N))
+    filtdims = ntuple(d->imfilter_inplace!(r, fill(1.0, inds[d]), kernels[d], border), Val(N))
```
timholy: Since this is inplace I think switching to 1.0 would be OK. Is this overflow prevention or something else?

appveyor.yml
```diff
@@ -1,9 +1,17 @@
 environment:
   matrix:
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x86/0.6/julia-0.6-latest-win32.exe"
-  - JULIA_URL: "https://julialang-s3.julialang.org/bin/winnt/x64/0.6/julia-0.6-latest-win64.exe"
-  - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x86/julia-latest-win32.exe"
-  - JULIA_URL: "https://julialangnightlies-s3.julialang.org/bin/winnt/x64/julia-latest-win64.exe"
+  - julia_version: 0.7
```
timholy: Should add 1.0 if you make any other changes (I haven't finished reading yet). If no changes are needed, let's merge and add that later.

test/gradient.jl
```diff
@@ -55,7 +55,20 @@ using Test
...
+
+# begin substitute
+            flag = false
+            try
+                res = @inferred(imgradients(img, kernelfunc, Inner()))
+            catch
+                flag = true
+            end
+            @test_broken flag == false
+# end substitute
```
timholy: Ideally we'd get this and the one below fixed. May not be vital though if it slows down merging any more.

src/imfilter.jl
RalphAS: IIRC there was an "inexact" error for one of the tests without this.

appveyor.yml
RalphAS: v1.0 is choking on at least one dependency, so I suggest updating CI in the next round.

PR #65, opened 2018-07-13
Latest attempt to get this working on 0.7

timholy:
This incorporates #41 and #64. As I am rushing to leave town for two weeks, most likely without any internet whatsoever, this is quick.

Most of the tests are working on 0.7. You need a fresh build of Julia and you also want https://github.com/JuliaLang/julia/pull/28100 and https://github.com/JuliaLang/julia/pull/28084. While I am gone I would urge members of the Images community to keep trying to get the JuliaImages packages working on 0.7: porting these has shaken out something like a dozen critical bugs in Base Julia and its stdlibs. Having put so much work into Julia and JuliaImages so far, it would be something of a shame if 1.0 got released and we couldn't really run JuliaImages packages.

Discourse advertised a slack channel on Fridays to help people port packages. Good luck!

PR #68, opened 2018-07-22
CartesianRange fixes

ajdunlap:
When I try to run this package with the latest Julia 0.7 beta I get errors of the form

```
ERROR: LoadError: LoadError: TypeError: in apply_type, in Vararg count, expected Int64, got Type{CartesianIndex{0}}
```

during precompilation. With these changes I no longer get the errors. But I don't know much about this so this could be totally the wrong thing to do.


ajdunlap:
Sorry, I see that these changes have been made in #65 .

PR #73, opened 2018-08-18
fix remaining depwarn

Evizero:
This depwarn was caught by the Images.jl tests (but not the tests of this package I think)

```
┌ Warning: `CartesianIndices(inds::Vararg{AbstractUnitRange{Int}, N}) where N` is deprecated, use `CartesianIndices(inds)` instead.
│   caller = _imfilter_inplace_tuple! at imfilter.jl:1270 [inlined]
└ @ Core ~/.julia/dev/ImageFiltering/src/imfilter.jl:1270
```

### DIFF
--- a/src/imfilter.jl
+++ b/src/imfilter.jl
@@ -1271,7 +1271,7 @@ function _imfilter_inplace_tuple!(r, out, img, kernel, Rbegin, inds, Rend, borde
                              out,
                              out,
                              tail(kernel),
-                             CartesianIndices(Rbegin.indices..., ind),
+                             CartesianIndices((Rbegin.indices..., ind)),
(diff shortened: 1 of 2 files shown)


timholy:
Suggests our ImageFiltering tests didn't catch it. Can a test be added that would have caught it?

Evizero:
I added a small code snippet that executes the method and would have thus triggered the depwarn.

Its not much of a tests, but after 10 minutes of trying to find something convenient to test against I thought to myself "maybe thats good enough". (Its not exactly my domain)

timholy:
Release tagged. (Yes it's small, but presumably the Images tests will fail without it.)

PR #75, opened 2018-09-09
mapwindow arguments positional to keywords

jw3126:
Now that Julia has fast keyword arguments, I think we should make use of them to shorten some signatures. This PR is for `mapwindow` and `mapwindow!`. 

I also looked at `imfilter` and `imfilter!`. These would in principle profit from some keyword arguments as well. However
* I personally do not use the more advanced signatures in `imfilter` and `imfilter!`
* There is a lot of "dispatch code" for `imfilter` and `imfilter!` already, which I am not familiar with. While I could blindly add yet another layer, I think it would be much better if a power user of these functions does this properly.
So I prefer not to touch `imfilter` and `imfilter!`, at least for this PR.

### DIFF
--- a/src/mapwindow.jl
+++ b/src/mapwindow.jl
@@ -9,7 +9,7 @@ using Statistics
 export mapwindow, mapwindow!
 
 """
-    mapwindow(f, img, window, [border="replicate"], [imginds=axes(img)]) -> imgf
+    mapwindow(f, img, window; [border="replicate"], [indices=axes(img)]) -> imgf
 
 Apply `f` to sliding windows of `img`, with window size or axes
 specified by `window`. For example, `mapwindow(median!, img, window)`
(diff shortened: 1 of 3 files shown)


timholy:
:+1: I just have one question: did you by any chance run the old tests on the modified code to make sure the deprecations you so wisely added actually work? If so, we can merge this ASAP.

> So I prefer not to touch imfilter and imfilter!, at least for this PR.

Absolutely fine, I sympathize completely. The dispatch hierarchy was very tricky to get right.

src/mapwindow.jl
```diff
@@ -438,5 +438,7 @@ end
 default_shape(::Any) = identity
 default_shape(::typeof(median_fast!)) = vec
 
+@deprecate mapwindow(f, img, window, border, indices=axes(img)) mapwindow(f,img,window,border=border,indices=imginds)
```
Evizero: should this be `imginds=default_imginds(img)`? This doesn't look like it works. I expect this to throw an `UndefVarError`

jw3126:
> +1 I just have one question: did you by any chance run the old tests on the modified code to make sure the deprecations you so wisely added actually work?

Yes, but then I changed something and did not run them again :smile:. Anyway now the deprecations pass the old tests.

timholy:
Thanks!

PR #81, opened 2018-10-12
add and expose `psf2otf` and `otf2psf`

johnnychen94:
Add commonly used functions `psf2otf` and `otf2psf`

Note:
The implementation mainly comes from the original MATLAB codes, e.g., `edit psf2otf`. The original code handles roundoff error, which in Julia there's no such need **(I guess? )**, so I skipped this part.

Reference:
http://www.mathworks.com/help/images/ref/psf2otf.html
http://www.mathworks.com/help/images/ref/otf2psf.html


johnnychen94:
Yeah, understood. I wasn’t aware of the license issue.

timholy:
I'm sure you weren't, many people aren't. I know you put some effort into this, sorry not to be able to use it.

Issue #98, opened 2019-04-24
Lazy padarray?


jw3126:
I think it would be useful to have a lazy variant of `padarray`. E.g. a thin wrapper around the array that computes border values in `getindex` calls. 
I have the feeling that this thing exists already somewhere? If not, is there interest of having that functionality here?

timholy:
There's [PaddedViews](https://github.com/JuliaArrays/PaddedViews.jl). But it's not as flexible. Yes, it probably makes sense to do that.

jw3126:
Solved by #99

PR #99, opened 2019-04-24
add BorderArray

jw3126:
#98

### DIFF
--- a/src/border.jl
+++ b/src/border.jl
@@ -236,8 +236,12 @@ function padfft(indk::AbstractUnitRange, l::Integer)
     range(first(indk), length=nextprod([2,3], l+lk)-l+1)
 end
 
+function error_bad_padding_size(inner, border)
+    ArgumentError("$border lacks the proper padding sizes for an array with $(ndims(inner)) dimensions")
+end
+
 function padindices(img::AbstractArray{<:Any,N}, border::Pad) where N
-    throw(ArgumentError("$border lacks the proper padding sizes for an array with $(ndims(img)) dimensions"))
+    throw(error_bad_padding_size(img, border))
 end
 function padindices(img::AbstractArray{<:Any,N}, border::Pad{N}) where N
     _padindices(border, border.lo, axes(img), border.hi)
--- a/src/border.jl
+++ b/src/border.jl
@@ -652,13 +652,11 @@ The command `padarray(A,Fill(0,(1,1,1)))` yields
 ---
 
 """
-padarray(img::AbstractArray, border::Pad)  = padarray(eltype(img), img, border)
-function padarray(::Type{T}, img::AbstractArray, border::Pad) where T
-    inds = padindices(img, border)
-    # like img[inds...] except that we can control the element type
-    newinds = map(Base.axes1, inds)
-    dest = similar(img, T, newinds)
-    copydata!(dest, img, inds)
+padarray(img::AbstractArray, border::AbstractBorder)  = padarray(eltype(img), img, border)
+function padarray(::Type{T}, img::AbstractArray, border) where {T}
+    ba = BorderArray(img, border)
+    out = similar(ba, T)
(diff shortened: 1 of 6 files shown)


jw3126:
@timholy what do you think about the general design of `BorderArray`?
* I did not implement `setindex!` since it would be mutating border + inner points simultaneously
* What should happen with `padarray`? Should it stay as is? Should it return a `BorderArray`? Should it be removed? Personally I think it could be return a `BorderArray` and the current `padarray` should be spelled something like `full(padarray(...))` instead.

timholy:
This looks superficially awesome, but I haven't had bandwidth recently. Maybe I can review this tomorrow, sorry for the delay.

src/borderarray.jl
```diff
@@ -0,0 +1,170 @@
...
+        if !compatible_dimensions(arr, border)
+            throw(ArgumentError("$border lacks the proper padding sizes for an array with $(ndims(arr)) dimensions"))
+        end
+        A = typeof(arr)
+        B = typeof(border)
+        new{T,N,A,B}(arr, border)
+    end
+end
+
+# these are adaped from OffsetArrays
```
timholy:
```suggestion
# these are adapted from OffsetArrays
```

src/borderarray.jl
```diff
@@ -0,0 +1,170 @@
...
+    ci = CartesianIndex(inds)
+    o[ci]
+end
+
+@inline function Base.getindex(o::BorderArray, ci::CartesianIndex)
+    if checkbounds(Bool, o.inner, ci)
+        @inbounds o.inner[ci]
+    else
+        @boundscheck checkbounds(o, ci)
+        getindex_outer_inbounds(o.inner, o.border, ci)
```
timholy: Do we need to make sure that both paths return values of type `T`? AFAICT it's possible to do something like
```julia
v = BorderArray([1, 2], Fill(NaN, (1,)))
```
and then indexing `v` is non-inferrable.

src/border.jl
```diff
@@ -652,13 +652,11 @@ The command `padarray(A,Fill(0,(1,1,1)))` yields
...
-    inds = padindices(img, border)
-    # like img[inds...] except that we can control the element type
-    newinds = map(Base.axes1, inds)
-    dest = similar(img, T, newinds)
-    copydata!(dest, img, inds)
+padarray(img::AbstractArray, border::AbstractBorder)  = padarray(eltype(img), img, border)
+function padarray(::Type{T}, img::AbstractArray, border) where {T}
+    ba = BorderArray(img, border)
+    out = similar(ba, T)
+    copy!(out, ba)
```
timholy: Can you do some performance comparisons between the existing implementation and this one? Your performance optimizations for `_copy!` are encouraging, of course, but just want to check.

jw3126: Yes I did. The current implementation is basically stolen from `padarray` anyway. I tinkered a bit with making it simpler or using the generic `copy`, but sadly performance was not as good.

src/borderarray.jl
jw3126: Good point! Do you prefer to throw an error in the constructor or a `convert` in `getindex` if that happens?

timholy: If it's straightforward to do it in the constructor, that would be best.

timholy:
Once this passes tests, I approve merging. Thanks for this, fantastic work!

jw3126:
@timholy can we tag a new release?

timholy:
Sorry I didn't notice the question about tagging. Absolutely! There's nothing breaking here, right? So this could be 0.6.1?

I think you should be able tag yourself, if you like. Perhaps give it a try, just so we know whether you are as empowered as you deserve to be.

jw3126:
Yeah, there is no breaking change, at least no intended one :smile:. I will see if I can create a release.

PR #100, opened 2019-05-02
Add freqkernel and spacekernel

timholy:
This replaces #81. Comments welcome!


src/utils.jl
```diff
@@ -13,6 +13,71 @@ See also: [`imfilter`](@ref).
...
+function, and known in some frameworks as `psf2otf`. If `rfft` is `true`, the fft for real-valued
+arrays (`rfft`) is returned instead and the first dimension size will be approximately half of `sz[1]`.
+
+`kern` should be zero-centered, i.e., `kern[0, 0]` should reference the center of your kernel.
+See [`centered`](@ref). Optionally specify the numeric type `T` (which must be one of the
+types supported by FFTW, either `Float32` or `Float64`).
+
+The inverse of `freqkernel` is [`spacekernel`](@ref).
+"""
+function freqkernel(::Type{T}, kern::AbstractArray, sz::Dims=size(kern); rfft=false) where T<:Union{Float32,Float64}
```
johnnychen94: Since `T` is float type, will it be reasonable to use `rfft` by default? Or is there a need to use `fft` rather than `rfft` here?

timholy: Great question. While it's not done here, one could implement `freqkernel` for complex-valued kernels. If we make `rfft` the default it will have to be a different API, but with `fft` the default then the generalization is basically trivial.

timholy:
I'll merge this soon, barring any further feedback.

PR #104, opened 2019-05-28
correct the behavior of centered for OffsetArray

johnnychen94:
This patch generalizes `centered(A::AbstractArray)` by not assuming indexes of A start at 1.

A side effect of this patch is to corrects the behavior of `imfilter` with accident usage of `centered(::OffsetArray)`

### Before:
```julia
julia> x = testimage("cameraman")

julia> k = KernelFactors.gaussian(1.5, 11)
OffsetArray(::Array{Float64,1}, -5:5) with eltype Float64 with indices -5:5:

julia> a = imfilter(x, kernelfactors((k, k)))

julia> b = imfilter(x, kernelfactors((centered(k), centered(k))))

julia> a - b
```

image

### After:
```julia
julia> gray.(sum(abs.(a-b)))
0.0
```

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -12,7 +12,6 @@ arbitrary axes.
 See also: [`imfilter`](@ref).
 """
 function centered(A::AbstractArray)
-    all(isodd.(size(A))) || error("entries must be odd, got $A")
     offsetted = first.(axes(A)) .- 1
     total_offsets = .-((size(A) .+ 1 ) .÷ 2)
     OffsetArray(A, total_offsets .- offsetted)
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -11,7 +11,12 @@ arbitrary axes.
 
 See also: [`imfilter`](@ref).
 """
-centered(A::AbstractArray) = OffsetArray(A, map(n->-((n+1)>>1), size(A)))
+function centered(A::AbstractArray)
+    all(isodd.(size(A))) || error("entries must be odd, got $A")
+    offsetted = first.(axes(A)) .- 1
+    total_offsets = .-((size(A) .+ 1 ) .÷ 2)
+    OffsetArray(A, total_offsets .- offsetted)
+end
 
 """
     kernfft = freqkernel([T::Type], kern, sz=size(kern); rfft=false)
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "ImageFiltering"
 uuid = "6a3955dd-da59-5b1f-98d4-e7296123deb5"
-version = "0.6.2"
+version = "0.6.3"
 
 [deps]
 CatIndices = "aafaddc9-749c-510e-ac4f-586e18779b91"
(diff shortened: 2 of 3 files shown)


zygmuntszpak:
In summary, this pull-request ensures that if you call `centered` on an `OffsetArray` it will adjust the axes of the _original_ data array that was being wrapped.

For example, 
```julia
axes(centered(collect(1:5))) # (Base.IdentityUnitRange(-2:2),)

axes(OffsetArray(collect(1:5),-1)) # (Base.IdentityUnitRange(0:4),)

axes(centered(OffsetArray(collect(1:5),-1)))  # (Base.IdentityUnitRange(-3:1),)  (Without this pull-request)

axes(centered(OffsetArray(collect(1:5),-1)))  # (Base.IdentityUnitRange(-2:2),)  (Thanks to this pull-request)

```

@timholy Are you OK with this new behaviour?

src/utils.jl
```diff
@@ -12,6 +12,7 @@ arbitrary axes.
 See also: [`imfilter`](@ref).
 """
 centered(A::AbstractArray) = OffsetArray(A, map(n->-((n+1)>>1), size(A)))
```
timholy: It's better to solve this in terms of *values* than in terms of *types*. For example, in my lab's https://github.com/HolyLab/CenterIndexedArrays.jl it defines an array type that is centered on 0, yet these are not `OffsetArray`s. Likewise, if you have an OffsetArray wrapping some other array type whose indexes don't start with 1, your fix will give the wrong answer.

So rather than defining the new method below, the best option would be to redefine this method so that it doesn't implicitly assume that `A`'s indexes start with 1.

johnnychen94: Is there any efficient way to get the latent Array for general arrays (e.g., ReshapedArray, SubArray, OffsetArray) compared to reconstructing a new one: `eltype(A).(A)` ?

`parent(::OffsetArray)` works for this purpose but `parent(::SubArray)` and `parent(reshape(1:9, 3, 3))` doesn't.

johnnychen94: I'm having a tendency to dispatches everything, without being aware of such an explicit solution 😂 Thanks for pointing it out.

src/utils.jl
```diff
@@ -11,7 +11,12 @@ arbitrary axes.
 
 See also: [`imfilter`](@ref).
 """
-centered(A::AbstractArray) = OffsetArray(A, map(n->-((n+1)>>1), size(A)))
+function centered(A::AbstractArray)
+    all(isodd.(size(A))) || error("entries must be odd, got $A")
```
timholy: This didn't used to be a requirement, though its behavior wasn't documented (which is not good):
```julia
julia> centered(1:4)
1:4 with indices -1:2
```

I think if you delete this check it's good to go. (Also for future purposes, that error message is confusing: "entries must be odd" sounds like the values of A, and that's compounded by the fact that you print A itself rather than its axes.)

timholy:
Thanks for doing this!

PR #106, opened 2019-07-05
[WIP] update benchmark script

johnnychen94:
The current benchmark script is broken, so I redesign it. Hopefully, we can catch a bunch of performance bottleneck with it, and decide where the future efforts should be put on.

A comprehensive benchmark on `ImageFiltering` requires more utils and benchmark items, but I'd like to get some early feedback on this.

I plan to make the benchmark script a little bit redundancy so that we can get more performance information about `ImageFiltering`. However, this requires quite a long-running time on the local machine (e.g., ~4mins in i7-6700K CPU @ 4.00GHz in the first commit)

TODO:

- [ ] `imfilter`
  - [ ] algorithm specification
  - [ ] computational resource specification
  - [ ] kernel: separable or not
- [ ] `mapwindow`
- [ ] visualization of benchmark results
- [ ] easy-to-use benchmarking pipeline script

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -1,4 +1,8 @@
 using ImageFiltering
-using PkgBenchmark
+using TestImages, ImageTransformations
+using BenchmarkTools
 
-include("mapwindow.jl")
+const SUITE = BenchmarkGroup(["ImageFIltering"])
+
+include("utils.jl")
+include("imfilter.jl")
--- a/benchmark/imfilter.jl
+++ b/benchmark/imfilter.jl
@@ -1,0 +1,39 @@
+SUITE["imfilter"] = BenchmarkGroup(["imfilter"])
+SUITE["imfilter_naive"] = BenchmarkGroup(["imfilter", "naive implementation"])
+SUITE["imfilter_naive"]["CPU1"] = BenchmarkGroup(["CPU", "single thread"])
+SUITE["imfilter"]["CPU1"] = BenchmarkGroup(["CPU", "single thread"])
+# SUITE["imfilter"]["CPUThread"] = BenchmarkGroup(["CPU", "multi threads"])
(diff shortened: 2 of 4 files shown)


benchmark/benchmarks.jl
```diff
@@ -1,4 +1,8 @@
 using ImageFiltering
-using PkgBenchmark
+using TestImages, ImageTransformations
+using BenchmarkTools
 
-include("mapwindow.jl")
+const SUITE = BenchmarkGroup(["ImageFIltering"])
```
timholy:
```suggestion
const SUITE = BenchmarkGroup(["ImageFiltering"])
```

benchmark/imfilter.jl
```diff
@@ -0,0 +1,39 @@
...
+# SUITE["imfilter"]["CUDA"] = BenchmarkGroup(["GPU", "CUDA"])
+# SUITE["imfilter"]["ArrayFire"] = BenchmarkGroup(["GPU", "ArrayFire"])
+# SUITE["imfilter"]["OpenCL"] = BenchmarkGroup(["GPU", "OpenCL"])
+
+img_gray_256 = testimage("lena_gray_256")
+img_rgb_256 = testimage("lena_color_256")
+img_list_gray = [imresize(img_gray_256, ratio=ratio) for ratio in (0.5, 1.0, 2.0, 4.0)]
+img_list_rgb = [imresize(img_rgb_256, ratio=ratio) for ratio in (0.5, 1.0, 2.0, 4.0)]
+
+default_kernel = centered(ones((31, 31))./prod((31, 31)))
```
timholy: What led you to choose the sizes here? (Both for the kernel and the images in the lists.) If running time is a concern we could do everything with smaller images and/or kernels.

johnnychen94:
I completely forget about it. Close in favor of #158

Issue #105, opened 2019-06-12
mapwindow doesn't work with 4D arrays


ChantalJuntao:
````````````````````````````````````````
A = rand(5,5,5)
mapwindow(mean, A, [1,1,3])
``````````````````````````````````````````
seems to work fine

````````````````````````````````````````
A = rand(5,5,5,5)
mapwindow(mean, A, [1,1,1,3])
``````````````````````````````````````````
returns the following error

``````````````````````
ERROR: MethodError: no method matching ntuple(::getfield(ImageFiltering.MapWindow, Symbol("##25#26")), ::Type{Val{4}})Closest candidates are:
  ntuple(::F, ::Integer) where F at tuple.jl:136
  ntuple(::Any, ::Val{0}) at tuple.jl:158
  ntuple(::Any, ::Val{1}) at tuple.jl:159
  ...
Stacktrace:
 [1] docopy! at /home/chantal/.julia/packages/ImageFiltering/8fmJ4/src/mapwindow.jl:294 [inlined]
 [2] copy_win!(::Array{Float64,4}, ::Array{Float64,4}, ::CartesianIndex{4}, ::Pad{0}, ::CartesianIndex{4}) at /home/chantal/.julia/packages/ImageFiltering/8fmJ4/src/mapwindow.jl:287
 [3] make_buffer_values_realistic!(::Array{Float64,4}, ::Array{Float64,4}, ::NTuple{4,UnitRange{Int64}}, ::Pad{0}, ::NTuple{4,Base.OneTo{Int64}}) at /home/chantal/.julia/packages/ImageFiltering/8fmJ4/src/mapwindow.jl:240
 [4] compute_output_eltype(::typeof(mean), ::Array{Float64,4}, ::NTuple{4,UnitRange{Int64}}, ::Pad{0}, ::NTuple{4,Base.OneTo{Int64}}) at /home/chantal/.julia/packages/ImageFiltering/8fmJ4/src/mapwindow.jl:228
 [5] allocate_output(::Function, ::Array{Float64,4}, ::NTuple{4,UnitRange{Int64}}, ::Pad{0}, ::NTuple{4,Base.OneTo{Int64}}) at /home/chantal/.julia/packages/ImageFiltering/8fmJ4/src/mapwindow.jl:214
[...]

PR #107, opened 2019-07-18
fixed issue #105 dimension deprecation

ChantalJuntao:
Thanks @jw3126  for the suggestion


timholy:
Test failure is only on nightly and seems to indicate we need to add Random to `Project.toml`, `[extras]`. Thanks @ChantalJuntao!

jw3126:
> 
> 
> Thanks @jw3126 for the suggestion

@ChantalJuntao your welcome. Thanks for the contribution!

PR #115, opened 2019-11-06
Improve the performance of mapwindow(extrema, ...), esp for large windows

timholy:


src/mapwindow.jl
```diff
@@ -378,10 +378,23 @@ extrema_filter(A::AbstractArray{T,N}, window::Integer) where {T,N} = extrema_fil
...
+            n = w1>>1
+            cache = CircularDeque{typeof(a)}(n)
+            for i = 1:n
+                push!(cache, a)
+            end
+            _extrema_filter1!(A, w1, cache)
+        end
+    end
+    if ndims(A) > 1
+        _extrema_filter!(permutedims(A, [2:ndims(A);1]), w...)
```
johnnychen94: how about `permuteddimsview` here? Is that still useful?

src/mapwindow.jl
```diff
@@ -378,10 +378,23 @@ extrema_filter(A::AbstractArray{T,N}, window::Integer) where {T,N} = extrema_fil
...
-        _extrema_filter1!(A, w1, cache)
+        if w1 <= 20
+            cache = ntuple(i->a, w1>>1)
+            _extrema_filter1!(A, w1, cache)
+        else
+            n = w1>>1
+            cache = CircularDeque{typeof(a)}(n)
+            for i = 1:n
+                push!(cache, a)
+            end
```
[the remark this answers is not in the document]
timholy: Didn't check recently, but my code comment above suggests I had tried it at one point.

src/mapwindow.jl
timholy: It might be. This could have been written before we had that capability. But I didn't test. Sometimes it's better to copy if you're going to access the same values repeatedly. But there doesn't seem to be a lot of that here, so perhaps the permuteddimsview would be worth trying.

Issue #137, opened 2019-12-25
bug in similar() for BorderArrays


Cody-G:
I came across this when trying to compose BorderArrays to set a different fill value for the front and back of the array.  Code to demonstrate:

```julia
using ImageFiltering

b_front = BorderArray(zeros(10), Fill(1.0, (3,), (0,)))
b_both = BorderArray(b_front, Fill(1.0, (0,), (5,)))
b_sim = similar(b_both)
axes(b_sim) == axes(b_both)  # is false, should be true
```

I'll make a PR shortly

jw3126:
Fixed by #138

PR #138, opened 2019-12-26
fix similar() and axes() for BorderArray

Cody-G:
Addresses #137

### DIFF
--- a/src/borderarray.jl
+++ b/src/borderarray.jl
@@ -74,7 +74,7 @@ function Base.similar(A::BorderArray, ::Type{T}, dims::Dims) where T
 end
 const OffsetAxis = Union{Integer, UnitRange, Base.OneTo, IdentityUnitRange}
 function Base.similar(A::BorderArray, ::Type{T}, inds::Tuple{OffsetAxis,Vararg{OffsetAxis}}) where T
-    similar(A.inner, T, axes(A))
+    similar(A.inner, T, inds)
 end
 
 @inline function Base.axes(o::BorderArray)
@@ -91,7 +91,7 @@ end
 
 @inline function _outeraxes(arr, border)
     map(axes(arr), border.lo, border.hi) do r, lo, hi
-        (first(r)-lo):(last(r)+hi)
+        IdentityUnitRange((first(r)-lo):(last(r)+hi))
     end
 end
 
--- a/test/border.jl
+++ b/test/border.jl
@@ -1,4 +1,6 @@
 using ImageFiltering, OffsetArrays, ImageCore, Random
+import OffsetArrays.IdentityUnitRange
+
 using Test
 
 @testset "Border" begin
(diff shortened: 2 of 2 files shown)


Cody-G:
I also rolled in a fix for #139

src/borderarray.jl
```diff
@@ -91,7 +91,7 @@ end
 
 @inline function _outeraxes(arr, border)
     map(axes(arr), border.lo, border.hi) do r, lo, hi
-        (first(r)-lo):(last(r)+hi)
+        Base.IdentityUnitRange((first(r)-lo):(last(r)+hi))
```
johnnychen94: I'm not familiar with the code base thus I don't know if this change is expected.

But to keep compatibility to julia v1.0, you could add a custom implementation
```julia
if VERSION < v"1.1"
    IdentityUnitRange = Base.IdentityUnitRange
else
    IdentityUnitRange = # self-defined
end
```

test/border.jl
```diff
@@ -270,6 +270,13 @@ using Test
         a0 = reshape([1])  # 0-dimensional
         @test ImageFiltering.accumulate_padding((), a0) == ()
         @test ImageFiltering.accumulate_padding((0:1, -1:1), a0) == (0:1, -1:1)
+        # ImageFiltering #137
+        b_front = BorderArray(zeros(10), Fill(1.0, (3,), (0,)))
+        b_both = BorderArray(b_front, Fill(1.0, (0,), (5,)))
+        b_sim = similar(b_both)
+        @test axes(b_sim) == axes(b_both)
```
jw3126: Can you maybe also add a few very explicit `axes` tests? Like
`@test axes(b) === (IdentityUnitRange(...)),)`?

src/borderarray.jl
```diff
@@ -74,7 +74,7 @@ function Base.similar(A::BorderArray, ::Type{T}, dims::Dims) where T
 end
 const OffsetAxis = Union{Integer, UnitRange, Base.OneTo, IdentityUnitRange}
 function Base.similar(A::BorderArray, ::Type{T}, inds::Tuple{OffsetAxis,Vararg{OffsetAxis}}) where T
```
jw3126: Can you add tests for this method specifically?

src/borderarray.jl
timholy: The best portability option is `OffsetArrays.IdentityUnitRange` which is defined here: https://github.com/JuliaArrays/OffsetArrays.jl/blob/e22f56ed7f80048ebe9662eeae7dc2891ecf0b8d/src/OffsetArrays.jl#L6-L10

test/border.jl
timholy: Or rather than testing the type, maybe best would be to test with `==`, which should test both the values of the axes and the axes-of-the-axes.

Cody-G: I think I'm already testing with `==` rather than testing the type?  So I'm not sure what you mean here.

timholy: Don't just test that they are the same, test the actual value produced. E.g., if you expect `f(x)` to return 1, testing `f(x) == f(y)` will not catch certain classes of bugs.

Cody-G: Ah got it, thanks!

Cody-G: > Can you maybe also add a few very explicit `axes` tests? Like
> `@test axes(b) === (IdentityUnitRange(...)),)`?

I added that test, just let me know if there are others you'd like added

Cody-G:
Thanks for the reviews!  I've addressed the comments and rebased.  Just let me know if there are additional ways that I should be testing `axes()` and `similar()`, I'm not sure if I covered everything.

src/borderarray.jl
timholy:
```suggestion
        OffsetArrays.IdentityUnitRange((first(r)-lo):(last(r)+hi))
```

Cody-G: Oh shoot, I thought I switched that, sorry.  Will fix now

Cody-G: Should be fixed with the latest push (`OffsetArrays.IdentityUnitRange` was already imported so I didn't need any explicit prefix)

timholy:
Thanks, @Cody-G!

Issue #84, opened 2018-11-13
Documentation of Pad inconsistent


rapus95:
As far as i can tell, the examples for `:reflect` and `:symmetric` in [padarray](https://juliaimages.github.io/latest/function_reference.html#ImageFiltering.padarray) are contrary to the definitions  of the symbols in [Pad](https://juliaimages.github.io/latest/function_reference.html#ImageFiltering.Pad) in that they are switched

timholy:
Can you be a bit more explicit? I'm not seeing what you're seeing.

rapus95:
In the `:symmetric` example of padarray the borderline is duplicated outside aswell though for the `:symmetric` example in Pad it is not (border is `e d c b | a b c d ...`)

timholy:
Got it now. Yes, you are right. Want to [submit a pull request fixing it?](https://help.github.com/articles/editing-files-in-your-repository/)

rapus95:
First I'd need to know which one is the correct one, then I can swap headings.

timholy:
Oh, I see. The ultimate source is usually (though not always) the code itself:

```julia
julia> using ImageFiltering

julia> a = [1,2,3,4]
4-element Array{Int64,1}:
 1
 2
 3
 4

julia> padarray(a, Pad(:symmetric, 2))
OffsetArray(::Array{Int64,1}, -1:6) with eltype Int64 with indices -1:6:
 2
 1
 1
 2
 3
 4
 4
 3

julia> padarray(a, Pad(:reflect, 2))
OffsetArray(::Array{Int64,1}, -1:6) with eltype Int64 with indices -1:6:                                                                                                                                                                     
 3                                                                                                                                                                                                                                           
 2                                                                                                                                                                                                                                           
 1                                                                                                                                                                                                                                           
 2                                                                                                                                                                                                                                           
 3                                                                                                                                                                                                                                           
 4                                                                                                                                                                                                                                           
 3                                                                                                                                                                                                                                           
 2         
```

PR #141, opened 2020-01-09
Swap symmetric and reflect in doc

innerlee:
Fix https://github.com/JuliaImages/ImageFiltering.jl/issues/84


jw3126:
Can you add a doctest, like https://github.com/JuliaImages/ImageFiltering.jl/issues/84#issuecomment-438362454

innerlee:
There is no doctest in this repo. Maybe you mean a test?

jw3126:
Oh wow it seems you are right, I can't spot any doctests. But I guess that is because this package is older then `Documenter.jl` reasons and it is fine to add them @timholy ?

timholy:
Yes, it would be great to add doctests. And indeed this package predates Documenter.

PR #150, opened 2020-01-24
setup BenchmarkCI (changes on top of #148)

tkf:
changes on top of #148


jw3126:
Shall we merge it?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,25 @@
...
+      JULIA_DEBUG: BenchmarkCI
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@latest
+        with:
+          version: 1.3
+      - name: Install dependencies
+        run: julia -e 'using Pkg; pkg"add PkgBenchmark https://github.com/tkf/BenchmarkCI.jl"'
+      - name: Run benchmarks
+        run: julia -e 'using BenchmarkCI; BenchmarkCI.judge(baseline = "HEAD")'
```
jw3126:
```suggestion
        run: julia -e 'using BenchmarkCI; BenchmarkCI.judge()'
```

johnnychen94:
I'll do that when https://github.com/JuliaRegistries/General/pull/8348 is merged

.github/workflows/benchmark.yml
johnnychen94: Do we need to commit this? @tkf

tkf: Yes, it was for build passing even when origin/master (default `baseline`) is not ready for benchmark.

PR #151, opened 2020-01-27
imfilter docs: reorg with "Extended help"

timholy:
The idea is to prepare for https://github.com/JuliaLang/julia/pull/34226, which will appear in Julia 1.5. This PR is essentially a demo of how I'd think of organizing the help: a relatively brief initial section that is somewhere between an explanation and a reminder, i.e., sufficient for someone who already knows what she is doing. Then the full details follow.


timholy:
Let's use this PR to hash out some kind of consensus format and then merge. I'd propose we leave the conversion of the remaining docstrings as a good first issue for newcomers (I'm bad about coming up with simple tasks for others to get their feet wet). See #153.

Issue #152, opened 2020-01-27
Write benchmarks


timholy:
We could use many more benchmarks in the `benchmark` folder. If you're a newcomer, why is this an important contribution? There are surely ways the performance of this package could be improved, and having benchmarks makes it much easier for the algorithm developers to test the consequences of their changes (#154 is one of many potential examples).

timholy:
Closed by #158

Issue #142, opened 2020-01-13
TODOs


timholy:
There seem to be a number of people who are interested in making improvements in this package. To facilitate their explorations, I thought I'd list some things I know need fixing:

- [ ] Write more [benchmarks](https://github.com/JuliaImages/ImageFiltering.jl/tree/master/benchmark). These will be useful as any other changes to the package are made.
- [ ] Update usage of [OffsetArrays](https://github.com/JuliaArrays/OffsetArrays.jl): this package was developed before OffsetArrays was particularly efficient, and it contains *ugly* workarounds to strip off the OffsetArray wrapper before performing the operation. Get rid of these workarounds, and document/fix any resulting performance changes. (There might be performance *improvements* as well as a risk of regressions.) Be aware of https://github.com/JuliaArrays/OffsetArrays.jl/pull/90 which should arrive fairly soon.
- [ ] Fix inefficiencies of [TiledIteration](https://github.com/JuliaArrays/TiledIteration.jl). That package is potentially a huge performance booster, but https://stackoverflow.com/questions/47590839/unexpected-memory-allocation-when-using-array-views-julia/47607539#47607539 gets in the way. If the wrapper creation and usage can be inlined into a single caller, then the compiler will elide the wrapper creation. Also be aware of false sharing and the potential performance problems it can cause (good thread: https://discourse.julialang.org/t/parallel-reductions/30180)
- [ ] Look into whether the boundary conditions can be implemented more generally as views, perhaps moving more functionality to [PaddedViews](https://github.com/JuliaArrays/PaddedViews.jl).
- [ ] Check whether the threading implementation should be updated for Julia 1.3's new threading model
- [ ] Implement threading for non-separable FIR filtering. Currently it's implemented only for separable kernels, because that's what I use most.
- [ ] GPUs...
- [ ] Consider splitting the package into `ArrayFiltering`, keeping anything color-related in `ImageFiltering` (which would become a thin wrapper around `ArrayFiltering`). Lots of people who *don't* do image processing are turned off by the "Image" in the name; it's not rational, but it's been brought up again and again, so it's time to stop wishing others would change and just change ourselves.

Tokazama:
The last one seems really difficult and could lead to a lot of fruitless effort. For it to be successful I think we'd need active participation from the wider non JuliaImages community.

timholy:
[...]

PR #154, opened 2020-01-27
Remove unfortunate OffsetArrays specializations

timholy:
This addresses "Update usage of OffsetArrays" in #142. It wasn't as ugly as I feared, but a nice cleanup nonetheless.

Do not merge this until we get benchmarks that let us assess the consequences of this change. (#152)


timholy:
Bummer. Looks like they are still useful. I will have to dig deeper to figure out why.

timholy:
Bizarre. I couldn't replicate the performance difference on my laptop for a couple of test cases. Let's see if the rebase somehow magically fixed something. (Seems unlikely, but still...)

PR #155, opened 2020-02-04
add sliding_window

jw3126:
My main motivation was to have multi argument mapwindow. Now you can do things like
```julia
map(f, sliding_window(arr1, window1), sliding_window(arr2, window2), arr3)
```
etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ os:
   - osx
   - windows
 julia:
+  - 1.0
   - 1.2
   - 1.4
   - nightly
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,10 @@ os:
   - osx
   - windows
 julia:
-  - 1.0
-  - 1
+  - 1.1
+  - 1.2
+  - 1.3
+  - 1.4
   - nightly
 notifications:
   email: false
@@ -16,7 +18,7 @@ notifications:
 jobs:
   include:
     - stage: "Documentation"
-      julia: 1.0
+      julia: 1.3
       os: linux
       script:
         - julia --project=docs/ -e 'using Pkg; Pkg.develop(PackageSpec(path=pwd()));
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,7 @@ os:
   - osx
   - windows
 julia:
-  - 1.1
   - 1.2
-  - 1.3
   - 1.4
   - nightly
 notifications:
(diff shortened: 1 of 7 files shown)


jw3126:
Any ideas what the CI fail is about? It only happens on `1.0`, there I get `unreachable reached`.

timholy:
One option would be to make this available only on 1.x, where x is the lowest minor release that does support it. If you decide to go that way, currently our travis testing only does 1.0, 1=latest stable release, and nightly, so you should add your 1.x to the travis script.

If you want to solve the problem that way let me know, then I will give this a review.

jw3126:
Yeah that sounds like a pragmatic solution.

timholy:
Another approach would be to make this particular function available only on Julia > 1 with a suitable version check, and the docstring marked with
```julia
!!! compat "Julia 1.1"
    sliding_window is only available on Julia 1.1 and higher.
```
I can go either way, but I slightly prefer being able to continue to support Julia 1.0 for as long as that is feasible.

jw3126:
> Another approach would be to make this particular function available only on Julia > 1 with a suitable version check, and the docstring marked with

Done

timholy:
I'll get to this as soon as a I can. If you haven't heard from me in a week, please ping again.

jw3126:
@timholy just a reminder to review this.

johnnychen94:
It occurs to me that some of the functionality here is overlapped with TiledIteration. Currently, that package only supports disjoin tiles, perhaps a better strategy is to enhance that package with sliding semantics?

```julia
julia> img1 = testimage("camera");

julia> img2 = testimage("mandril_gray");

julia> iter1 = TileIterator(axes(img1), (9, 9))
TileIterator((Base.OneTo(512), Base.OneTo(512)), (9, 9))

julia> iter2 = TileIterator(axes(img2), (9, 9))
TileIterator((Base.OneTo(512), Base.OneTo(512)), (9, 9))

julia> map(iter1, iter2) do R1, R2
           mse(view(img1, R1...), view(img2, R2...))
       end;
3249-element Array{Float32,1}:
...
```

ImageFiltering is becoming unnecessarily complex and hard to maintain or enhance. My vision on ImageFiltering is to not adding internal array types into this package, and instead, maintaining a more generic version as an independent package.

johnnychen94:
With https://github.com/JuliaArrays/TiledIteration.jl/pull/23 merged, I believe this is no longer wanted?

jw3126:
Note that JuliaArrays/TiledIteration.jl#23 is not quite enough to cover this PR. For one a unit stride strategy is missing. And then this one does border handling + tiling, while `TiledIteratation` is only about the tiling. Nevertheless we can close this PR. While much of the functionality of this PR is still desirable, I would choose to implement it more on top of TiledIteratation now.

PR #157, opened 2020-02-10
Don't use FFT when input has NaNs

yha:
This fixes filtering of floating-point arrays with `NaN` and large kernels.
According to benchmarks on my machine, the time to check for `NaN`s is up to about 0.5% of the `imfilter` time (for `img::Vector{Float64}` and kernel of length 31).


timholy:
Nice, thanks!

PR #147, opened 2020-01-21
Add Moffat Kernel

gerrygralton:
In reference to feature request #146. 
Please  let me know if anything isn't done how it should be. I'm keen to learn.

PR #159, opened 2020-03-06
Fix Moffat tests

gerrygralton:
Fixing Moffat tests as noted by @timholy in https://github.com/JuliaImages/ImageFiltering.jl/pull/147#discussion_r387512964 and https://github.com/JuliaImages/ImageFiltering.jl/pull/147#discussion_r387521100


timholy:
Thanks @gerrygralton!

PR #160, opened 2020-03-23
Fix mapwindow for offset arrays

yha:
The deleted assertions fail due to  JuliaArrays/OffsetArrays.jl#104. As a workaround, I tried
```
@assert (imgr .+ kmin)[idx] ⊆ fullimgr
@assert (imgr .+ kmax)[idx] ⊆ fullimgr
```
instead, but that fails due to JuliaLang/julia#35225 when `idx` turns out empty.

### DIFF
--- a/src/mapwindow.jl
+++ b/src/mapwindow.jl
@@ -153,28 +153,26 @@ end
 struct _IdentityTransformer <: _IndexTransformer end
 @inline Base.getindex(t::_IdentityTransformer, inds) = inds
 
-function _IndexTransformer(ranges)
-    stride = map(step, ranges)
-    offset1 = map(first, ranges)
-    offset = offset1 .- stride
+function _IndexTransformer(from_ranges::NTuple{N,AbstractUnitRange}, to_ranges) where {N}
+    stride = map(step, to_ranges)
+    offset = first.(to_ranges) .- first.(from_ranges) .* stride
     _AffineTransformer(offset, stride)
 end
 
-function _IndexTransformer(ranges::NTuple{N,Base.OneTo}) where {N}
-    _IdentityTransformer()
+function _IndexTransformer(from_ranges::NTuple{N,AbstractUnitRange}, to_ranges::NTuple{N,AbstractUnitRange}) where {N}
+    offset = first.(to_ranges) .- first.(from_ranges)
+    _OffsetTransformer(offset)
 end
 
-function _IndexTransformer(ranges::NTuple{N,AbstractUnitRange}) where {N}
-    offset1 = map(first, ranges)
-    offset = offset1 .- 1
-    _OffsetTransformer(offset)
+function _IndexTransformer(::NTuple{N,Base.OneTo}, ::NTuple{N,Base.OneTo}) where {N}
+    _IdentityTransformer()
 end
 
 compute_output_range(r::AbstractUnitRange) = r
 compute_output_range(r::AbstractRange) = Base.OneTo(length(r))
 
 function compute_output_indices(imginds)
-    ranges = map(compute_output_range, imginds)
+    ranges = map(i->compute_output_range(axes(i,1)), imginds)
     # Base.similar does not like if some but not all ranges are Base.OneTo
     homogenize(ranges)
 end
@@ -197,7 +195,7 @@ end
 function _indexof(r::AbstractRange, x)
     T = eltype(r)
     @assert x ∈ r
-    i = one(T) + T((x - first(r)) / step(r))
+    i = T(firstindex(r) + (x - first(r)) / step(r))
     @assert r[i] == x
     i
 end
@@ -210,6 +208,9 @@ function _indices_of_interiour_range(
     idx1 = _intersectionindices(fullimgr, kmin .+ imgr)
     idx2 = _intersectionindices(fullimgr, kmax .+ imgr)
     idx = intersect(idx1, idx2)
+    @assert imgr[idx] .+ kmin ⊆ fullimgr
+    @assert imgr[idx] .+ kmax ⊆ fullimgr
+    idx
 end
 
 function _indices_of_interiour_indices(fullimginds, imginds, kerinds)
@@ -255,13 +256,13 @@ function mapwindow_kernel!(f,
 
     @assert map(length, imginds) == map(length, axes(out))
 
-    indind_full = map(r -> Base.OneTo(length(r)), imginds)
+    indind_full = map(r -> axes(r,1), imginds)
     indind_inner = _indices_of_interiour_indices(axes(img), imginds, window)
     Rindind_full = CartesianIndices(indind_full)
     Rindind_inner = CartesianIndices(indind_inner)
 
-    outindtrafo = _IndexTransformer(axes(out))
-    imgindtrafo = _IndexTransformer(imginds)
+    outindtrafo = _IndexTransformer(indind_full, axes(out))
+    imgindtrafo = _IndexTransformer(indind_full, imginds)
 
     buf, bufrs = allocate_buffer(f, img, window)
     Rbuf = CartesianIndices(size(buf))
--- a/src/mapwindow.jl
+++ b/src/mapwindow.jl
@@ -73,8 +73,9 @@ function mapwindow(f, img, window; border="replicate",
               resolve_imginds(indices))
 end
 
-# a unit range `r` having `r == axes(r)` which keeps its axes with
-# broadcasting (e.g., `axes(r.+1) == axes(r)`), unlike Base.IdentityUnitRange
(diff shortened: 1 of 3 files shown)


johnnychen94:
I think the assertation makes sense and should not be removed. The fact that the tests pass is more concerning IMO...

With https://github.com/JuliaArrays/OffsetArrays.jl/pull/105 and the following patch, the assertation passes, but there're some new errors in the added tests.

https://github.com/JuliaImages/ImageFiltering.jl/blob/master/src/mapwindow.jl#L200

```diff
function _indexof(r::AbstractRange, x)
    T = eltype(r)
    @assert x ∈ r
-	i = ont(T) + T((x - first(r)) / step(r))
+   i = first(axes(r)[1]) + T((x - first(r)) / step(r))
    @assert r[i] == x
    i
end
```

yha:
Updated version, which should work assuming JuliaArrays/OffsetArrays.jl#105 (and includes the assertions).
(I believe the previous version without the assertions was also technically correct, but in a weird way, with methods not doing exactly what their name suggests for offset arrays, and relying subtly on JuliaArrays/OffsetArrays.jl#104).

This version has `axes(mapwindow(..., indices=2:7),1) == 1:6`, which is what the docs says it should do, rather than `2:7` as before. This is also more consistent the behavior of non-unit-ranges, e.g., `indices=2:2:8`. I changed one test as a result, which previously tested for behavior which contradict the docs.

test/mapwindow.jl
```diff
@@ -71,6 +72,19 @@ using ImageFiltering: IdentityUnitRange
...
+    # offsets
+    @testset "offsets" begin
+        arrays = [rand(5), rand(5,5), rand(5,5,5)]
+        @testset "offsets ($f, $offset, $window, $dim)" for f in (extrema,maximum,minimum),
+                                                            offset in -5:5,
+                                                            window in [1:2:9; [0:2,-2:0]],
+                                                            (dim,a) in enumerate(arrays)
+            offsets = ntuple(_->offset,dim)
+            windows = ntuple(_->window,dim)
+            @test OffsetArray(mapwindow(f,a,windows),offsets) == mapwindow(f,OffsetArray(a,offsets),windows)
```
jw3126, line 84: I think this is a good test and we should have it. Just asking, is this test different from behavior on master? Could also add a case where there is no handcrafted version for `f`. E.g. `f = x -> sum(x)`

test/mapwindow.jl
```diff
@@ -115,7 +129,7 @@ using ImageFiltering: IdentityUnitRange
         @test expected == @inferred mapwindow!(f,out,img,window,indices=imginds)
     end
     for (inds, kw) ∈ [((Base.OneTo(3),), (border="replicate", indices=2:2:7)),
-                        ((2:7,), (indices=2:7,)),
+                        (axes(2:7), (indices=2:7,)),
```
jw3126, line 132: I do not have an opinion on whether this is a bug fix or breaking change.

test/mapwindow.jl
yha: The behavior on master is an assertion error when the input array has non-zero offset. 
The original version of this PR simply removed these assertions and loosened some type checks, which happens to work "accidentally", but would stop working once JuliaArrays/OffsetArrays.jl#104 is fixed (e.g. JuliaArrays/OffsetArrays.jl#105 is merged).
The current version is the opposite: it depends on JuliaArrays/OffsetArrays.jl#105 to work.

Not sure what you mean about handcrafted version for `f`. AFAICT there is no specialized implementation of `mapwindow` for `minimum` or `maximum`.

yha:
> Basically looks good to me, see comments above.

Note this should probably be merged only after JuliaArrays/OffsetArrays.jl#105, with a corresponding `compat` entry.

test/mapwindow.jl
jw3126: Thanks for clarifying. I do not remember for which functions there is a specialized implementation. Even if there is none for `minimum/maximum` right now, it could potentially be added in future, so I believe adding a lambda f won't hurt.

johnnychen94:
Sorry that I was too busy these days to take care of this PR. I'll find some time to work on JuliaArrays/OffsetArrays.jl#105 this later.

yha:
Added tests for anonymous functions and various border styles. Also fixed the axes for `Inner()` which I apparently broke.
Caught some new failure cases which already fail in master. I'll open a new issue for those.

johnnychen94:
reopen to retrigger the test

johnnychen94:
The failure in nightly is because I retriggered the CI too quickly that PkgServer (enabled by default since Julia 1.5) didn't get updated, hence it doesn't found `OffsetArrays v1.1.0`.

There seem to be other issues in the nightly test: https://github.com/JuliaCI/NanosoldierReports/blob/master/pkgeval/by_date/2020-06/13/logs/ImageFiltering/1.6.0-DEV-6b2ffd3913.log

yha:
> There seem to be other issues in the nightly test: https://github.com/JuliaCI/NanosoldierReports/blob/master/pkgeval/by_date/2020-06/13/logs/ImageFiltering/1.6.0-DEV-6b2ffd3913.log

This seems to be before the fix in #174.

johnnychen94:
It seems to be only an CI issue, the test passes locally for nightly build (mac and Linux).

I believe this PR is ready for merge and it's being unfairly delayed for quite a long time, so I plan to merge this next Monday if there are no further comments.

timholy:
Let's try re-running CI

src/mapwindow.jl
```diff
@@ -72,14 +73,18 @@ function mapwindow(f, img, window; border="replicate",
               resolve_imginds(indices))
 end
 
+# a unit range `r` having `r == axes(r)` which keeps its axes with
+# broadcasting (e.g., `axes(r.+1) == axes(r)`), unlike Base.IdentityUnitRange
+self_offset(r::AbstractUnitRange) = OffsetArrays.IdOffsetRange(1:length(r), first(r)-one(eltype(r)))
```
timholy, line 78: I'm not exactly sure I understand what you're trying to do. It doesn't quite do what it claims in the comment, for example
```julia
julia> r2 = OffsetArrays.IdOffsetRange(5:10, -3)
2:7

julia> r3 = self_offset(r2)
2:7

julia> first(r3)
2

julia> first(axes(r2, 1))
-2
```
What's the underlying goal? Would you like to create a range that preserves the values and indices of an original `AbstractUnitRange` but which behaves as described under broadcasting?

Or is this a situation where we can essentially ignore certain axes? For instance, we might imagine ignoring the axes of `imginds`, although alternatively they might specify the axes of the output.

src/mapwindow.jl
```diff
@@ -190,14 +193,14 @@ function _intersectionindices(full::AbstractUnitRange, r::AbstractRange)
     else
         ret = _indexof(r,first(r_sub)):_indexof(r,last(r_sub))
     end
-    @assert intersect(full, r) == r[ret]
+    @assert r_sub == r[ret] || isempty(r_sub) && isempty(ret)
     ret
 end
 
 function _indexof(r::AbstractRange, x)
     T = eltype(r)
```
timholy, line 201: Interestingly, this should really be `eltype(axes(r, 1))`, but that's always `Int`.

test/mapwindow.jl
```diff
@@ -71,6 +71,33 @@ using ImageFiltering: IdentityUnitRange
...
+
+                if border == Inner() && winlen > n
+                    @test_throws DimensionMismatch mw(a)
+                    @test_throws DimensionMismatch mw(OffsetArray(a,offsets))
+                else
+                    @test OffsetArray(mw(a),offsets) == mw(OffsetArray(a,offsets))
+                end
+            end
+        end
+    end
```
timholy, line 99: Really elegant test! Fails big-time on master :smile:.

src/mapwindow.jl
yha: I suppose the comment is confusing because it uses the name `r` for the output of the function when it's also as the name of the function parameter. It should really say that `s = self_offset(r)` has the same values as `r` but has itself as indices: `collect(s) == collect(r)` and `axes(s) == (s,)` (in your example, `axes(r3) == (2:7,) == (r3,)`).

yha: Fixed the comment

yha:
Bump

johnnychen94:
Hmmm... I'll merge this in days if there are no further comments

johnnychen94:
Given that @timholy is quite busy these days for a second-round comment, I'm merging this.

@yha This is really a nice and tough job! I trust you know how things going internally. Also, the test looks great to me, so let's test this patch in the wild.

Sorry again for the long delay here

PR #165, opened 2020-04-18
2D Tophat (Disk) Kernel

mileslucas:
This adds a 2D "tophat" kernel. This is geometrically just a disk. I'm not an images guy, per se, so if there is a more apt name than `tophat`, please inform me! 

The two functions added to `Kernel` are 
* `tophat(r)` - default size is `2r` rounded up to the nearest odd integer. The cells whose distance from the center is less than or equal to `r` will be constant and all other will be zero. The constant will be such that the sum is 1. 
* `tophat(r, ls)` - the same as above, but allos user-input size `ls`. This will handle even and odd based shapes, and even mixes of both (eg `(6, 7)`. Again, the pixels whose distance from the center is less than or equal to `r` will be a normalized constant.

I've added a docstring and imported correctly under `ImageFiltering.jl`. I've added it to the `Kernel` module docstring, too. Unfortunately I'm not having a great time parsing the test suite, so I don't have any tests. Please let me know where I can test this.

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -442,7 +442,7 @@ function tophat(r::Real, sizes::Tuple{Integer, Integer})
             kern[idx] = amplitude
         end
     end
-    return kern
+    return kern ./ sum(kern) # renormalize 
 end
 
 """
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -13,6 +13,7 @@ dimensionality. The following kernels are supported:
   - `Laplacian`
   - `gabor`
   - `moffat`
+  - `tophat`
 
 See also: [`KernelFactors`](@ref).
 """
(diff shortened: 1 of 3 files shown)


mileslucas:
Bumping this

jw3126:
I think this can be tested in a `@testset "tophat"` in `test/2d.jl`. Is this `tophat` function coming from matlab or somehow standard? There are a lot of variantions on how to turn a circle into pixels. E.g. constant value strictly inside, value proportional to circle area intersected with pixel etc.

mileslucas:
This implementation is based off the skimage.draw.disk implementation which is used in the python library which inspired this PR.

src/kernel.jl
```diff
@@ -416,6 +417,34 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
     sum(x.^2)
 end
 
+"""
+    tophat(r) -> k
+    tophat(r, ls) -> k
```
jw3126, line 422: Rename `ls` to `size` and explain it?

jw3126: Also some doctests would be nice.

src/kernel.jl
```diff
@@ -416,6 +417,34 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
...
+    kern = zeros(typeof(amplitude), u, v)
+
+    cy, cx = u / 2 + 0.5, v / 2 + 0.5
+    for idx in CartesianIndices(axes(kern))
+        x, y = idx.I
+        if (x - cx)^2 + (y - cy)^2 ≤ r^2
+            kern[idx] = amplitude
+        end
+    end
+    return kern ./ sum(kern) # renormalize
```
jw3126, line 445: Why do you compute amplitude above, if you normalize it away here anyway?

PR #172, opened 2020-06-09
Generalize `NA` to take `na` predicate as argument (not just `isnan`)

yha:
Useful for filtering `SVector` or `GeometryTypes.Point` arrays with `NA(x->any(isnan,x))`.
May be also useful to do `NA(ismissing)` in the future, but `imfilter` with `missing`s doesn't work yet for other reasons.

I removed the line `out[nanflag] .= nan(T)`, which may be breaking (changes the behavior of plain `NA()`). I'm not sure why it was there; seems to me like a bug.

### DIFF
--- a/src/border.jl
+++ b/src/border.jl
@@ -705,44 +705,37 @@ struct Inner{N} <: AbstractBorder
     hi::Dims{N}
 end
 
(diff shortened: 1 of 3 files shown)


timholy:
Let's re-run this now on top of #174.

PR #185, opened 2020-09-23
add a demo for median filters

johnnychen94:
The purpose here is to provide an example to work on the remote page feature in DemoCards. (https://github.com/johnnychen94/DemoCards.jl/issues/72)

Adds a `median_filter.jl` example to explain how `mapwindow` is used in practice. (I think it answers the question in https://github.com/JuliaImages/ImageFiltering.jl/pull/14#issuecomment-694343995)

### DIFF
--- a/docs/demos/filters/median_filter.jl
+++ b/docs/demos/filters/median_filter.jl
@@ -1,0 +1,83 @@
+# ---
+# title: Median filters
+# cover: assets/median.gif
+# author: Johnny Chen
+# date: 2020-09-23
+# ---
+
+# With `median` filter as an example, this demo shows how you could construct a custom kernal and
+# pass it to [`mapwindow`](@ref) for common stencil operations.
+
+# `ImageFiltering` does not provide you an out-of-box median filter (e.g., `medfilt2` in MATLAB).
+# This is because with [`mapwindow`](@ref) function, it is quite trivial a process to build and tailor one
+# such for your own usage.
+
+using ImageFiltering, ImageCore, ImageShow # or you could just `using Images`
+using TestImages
+using Statistics
+
+img = float.(reshape(1:25, 5, 5))
+patch_size = (3, 3)
+imgm = mapwindow(median, img, patch_size)
+
+# `mapwindow` provides a high-level interface to loop over the image and call a function (`median`
+# in this demo) on each patch/window. That said, it does the following loop in a more efficient way.
+
+## For simplicity, border condition is not included here.
+imgm = zeros(axes(img))
+R = CartesianIndices(img)
+I_first, I_last = first(R), last(R)
+Δ = CartesianIndex(patch_size .÷ 2)
+for I in R
+    patch = max(I_first, I-Δ):min(I_last, I+Δ)
+    imgm[I] = median(img[patch])
+end
+imgm
+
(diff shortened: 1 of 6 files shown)


docs/demos/filters/median_filter.jl
```diff
@@ -0,0 +1,83 @@
...
+
+# `ImageFiltering` does not provide you an out-of-box median filter (e.g., `medfilt2` in MATLAB).
+# This is because with [`mapwindow`](@ref) function, it is quite trivial a process to build and tailor one
+# such for your own usage.
+
+using ImageFiltering, ImageCore, ImageShow # or you could just `using Images`
+using TestImages
+using Statistics
+
+img = float.(reshape(1:25, 5, 5))
```
timholy, line 19: Would it be better to use a testcase that's more obvious? The only places where the image changes in this example is along the edges, and you therefore have to know something about the default boundary conditions to make sense of the result.

What about something like
```julia
julia> img = [isodd(i) - isodd(j) for i = 1:5, j = 1:5]
5×5 Matrix{Int64}:
  0  1   0  1   0
 -1  0  -1  0  -1
  0  1   0  1   0
 -1  0  -1  0  -1
  0  1   0  1   0

julia> img[3, 3] = 1000
1000

julia> imgm = mapwindow(median, img, patch_size)
5×5 Matrix{Float64}:
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  1.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
 0.0  0.0  1.0  0.0  0.0
 0.0  0.0  0.0  0.0  0.0
```

docs/demos/filters/median_filter.jl
```diff
@@ -0,0 +1,83 @@
...
+patch_size = (3, 3)
+imgm = mapwindow(median, img, patch_size)
+
+# `mapwindow` provides a high-level interface to loop over the image and call a function (`median`
+# in this demo) on each patch/window. That said, it does the following loop in a more efficient way.
```
timholy, line 24:
```suggestion
# in this demo) on each patch/window. It's performing an operation that is nearly equivalent to the loop below, albeit more efficiently:
```

docs/demos/filters/median_filter.jl
```diff
@@ -0,0 +1,83 @@
...
+R = CartesianIndices(img)
+I_first, I_last = first(R), last(R)
+Δ = CartesianIndex(patch_size .÷ 2)
+for I in R
+    patch = max(I_first, I-Δ):min(I_last, I+Δ)
+    imgm[I] = median(img[patch])
+end
+imgm
+
+# As you can see, except for the borders, a hand-written loop works basically the same as `mapwindow`.
```
timholy, line 37:
```suggestion
# Compared to this hand-written loop, `mapwindow` offers additional flexibility in handling the borders and allows you to process only a subset of the windows. For some functions (e.g., `min`, `max`, and `extrema`), `mapwindow` uses a custom implementation that is far more efficient than naively computing the function "freshly" on each window.
```

timholy (commented):
Even though I know you were prompted by other needs, great demo and much appreciated!

johnnychen94:
Thanks for the suggestions, it looks much better!

Issue #189, opened 2020-09-25
demo: maxmin filter examples


johnnychen94:
Took a quick look at [1] and I feel this is worth a demonstration card.

References:

[1] Verbeek, P. W., Vrooman, H. A., & Van Vliet, L. J. (1988). [Low-level image processing by max-min filters](https://core.ac.uk/download/pdf/194053536.pdf). Signal Processing, 15(3), 249-258.

timholy:
Pretty cool. And `mapwindow` is really efficient thanks to the Lemire filters.

Dsantra92:
Can I work on this ?

johnnychen94:
Yes, of course!

We are plan to host hacktoberfest locally in JuliaCN during Oct 17-18, so if you're going to work on this, please submit an early PR to notify others.

Dsantra92:
@johnnychen94  is a pr within this week is okay?

Dsantra92:
@johnnychen94 @timholy  just a few set of questions.

1. Firstly I found 2 ways to find `minmax`, one was to use `extrema` and other was to use `maximum` on top of `minimum`.Both using `mapwindow`.I found `extrema` to be very efficient but I had to go with `minimum`+`maximum`  because extracting min or max filtered images of `extrema` also incurrs some cost, or am I missing an easy way to do so ?

2. I have done the maxmin, minmax, min and max filters.Are there any more implementations you would like to add ?

johnnychen94:
I'm surprised that `extrema` even works. The performance difference might be explained because `extrema` is written in a more efficient way while `maximum` is actually implemented as `mapreduce(identity, max, img)`. You could use `@which maximum(patch)` to check where it's defined.

I suggest still using `minimum` and `maximum` because this performance difference, if needed, should be fixed upstream.

FYI, `mapwindow!` can be used to reduce additional memory allocation when you have successive filters.

> I have done the maxmin, minmax, min and max filters.Are there any more implementations you would like to add ?

I don't have specific ideas in mind, if you have good ideas, feel free to add them.

I haven't check it yet, https://github.com/sairus7/MaxMinFilters.jl sounds like a promising functionality to be added into ImageFiltering.

johnnychen94:
Added in #190

timholy:
In this case `extrema` is just being used for dispatch---it's not really running `extrema` per se. The point is that the Lemire algorithm requires that both the min and max be extracted---it's integral to the computation itself.

[...]

PR #190, opened 2020-10-06
Add max min filter demo

Dsantra92:
#189 Added demo for max, min filters

### DIFF
--- a/docs/demos/filters/min_max_filter.jl
+++ b/docs/demos/filters/min_max_filter.jl
@@ -8,20 +8,20 @@
 # In this tutorial we see how can we can effectively use max and min filter to distinguish
 # between smooth and texture edges  in grayscale images.
 
-# We will use the [`mapwindow`](@ref) function in `ImageFiltering.jl` which provides a general 
-# functionality to apply any function to the window around each pixel. 
+# We will use the [`mapwindow`](@ref) function in `ImageFiltering.jl` which provides a general
+# functionality to apply any function to the window around each pixel.
 
 using ImageCore, ImageShow, ImageFiltering
 using TestImages
 
-img = Gray.(testimage("mandrill"))      #Original Image       
+img = Gray.(testimage("mandrill"))      # Original Image
 
-# We can use the `minimum` function to compute the minimum of the grayscale values in the given matrix,or array.
-# For example : 
-minimum([Gray(0.7),Gray(0.5),Gray(0.0)]) ##Should return Gray(0.0) i.e black.
+# We can use the `minimum` function to compute the minimum of the grayscale values in the given
+# matrix or array. For example: 
+minimum([Gray(0.7),Gray(0.5),Gray(0.0)]) # Should return Gray(0.0) i.e black.
 # 
 ## Using the `mapwindow` function, we create an image of the local minimum.
-## The mapwindow maps the given function over a moving window of given size.
+## `mapwindow` maps the given function over a moving window of given size.
 img_min = mapwindow(minimum, img, (3, 3))
 ## Similarly for maximum
 img_max = mapwindow(maximum, img, (3, 3))
@@ -29,28 +29,28 @@ img_max = mapwindow(maximum, img, (3, 3))
 img_max_min = mapwindow(maximum, img_min, (3, 3))
 ## The min(max) filter
 img_min_max = mapwindow(minimum, img_max, (3, 3))
-mosaicview(img_min, img_max, img_max_min, img_min_max; nrow = 1)
+mosaicview(img_min, img_max, img_max_min, img_min_max; nrow=1)
 
 # Now that we are done with the basic filtered images, we proceed to the next part
 # which is edge detection using these filters.
 
-# For edge detection, we need to define thresholds for our image.The threshold is an important tool 
-# to binarize a grayscale image.The threshold value for a given pixel practically
-# decides if the pixel visible or not in the output image.However, appyling a global threshold might not consider the 
-# variation of colors/brightness within the image.Thus we consider an adaptive type theresholding method
+# For edge detection, we need to define thresholds for our image. The threshold is an important tool
+# to binarize a grayscale image. The threshold value for a given pixel practically
+# decides if the pixel visible or not in the output image. However, appyling a global threshold might not consider the
+# variation of colors/brightness within the image. Thus we consider an adaptive type theresholding method
 # here using max/min results.
 
-# The max(min) and min(max) filter effectively follows the smooth edges of the image.Therefore,
-# their average also follows the smooth parts of the image.If we use this image as a threshold for the original
+# The max(min) and min(max) filter effectively follows the smooth edges of the image. Therefore,
+# their average also follows the smooth parts of the image. If we use this image as a threshold for the original
 # image, the smooth parts of the original image will get filtered out, leaving only the texture and/or noise behind.
-# So we can use  the average of `img_max_min` and `img_min_max` as the texture threshold.
+# So we can use the average of `img_max_min` and `img_min_max` as the texture threshold.
 
-# The average of min and max filters also gives us the smooth edges, but it also includes the noise in the image. 
-# So using average of `img_min` and `img_max` as a threshold  will yeild only texture.
+# The average of min and max filters also gives us the smooth edges, but it also includes the noise in the image.
+# So using average of `img_min` and `img_max` as a threshold will yeild only texture.
 
-img_texture_noise_threshold = (img_max_min + img_max_min)./ 2
-img_texture_threshold = (img_max + img_min)./2
-mosaicview(img_texture_noise_threshold, img_texture_threshold; nrow = 1)
+img_texture_noise_threshold = (img_max_min + img_max_min) ./ 2
+img_texture_threshold = (img_max + img_min) ./ 2
+mosaicview(img_texture_noise_threshold, img_texture_threshold; nrow=1)
 #
 ## The Dynamic Gist is obtained by subtracting the img_texture_threshold from the original image.
 ## The filtered image gives us the texture of the image.
@@ -58,21 +58,22 @@ img_dynamic_gist = img - img_texture_threshold
 ## The Texture Gist is obtained by subtracting img_texture_noise_threshold from the original image.
 ## The filtered image gives us the texture along with the noise of the image.
 img_texture_gist = img - img_texture_noise_threshold
-mosaicview(img_dynamic_gist, img_texture_gist; nrow = 1)
+mosaicview(img_dynamic_gist, img_texture_gist; nrow=1)
 #
 ## We extract out the smooth/ramp parts of the image.
-## The darker section of the image consist of the ramp edges.The brighter pixels are mostly noise.
+## The darker section of the image consist of the ramp edges. The brighter pixels are mostly noise.
 ramp = img_dynamic_gist - img_texture_gist
-## Filtered Out edges
+## Filtered-out edges
 edge = img_max - img_min
 ## Smoothed-out version of edge
-edge_smoothed = img_min_max- img_max_min
-mosaicview(img, ramp, edge, edge_smoothed; nrow = 2)
+edge_smoothed = img_min_max - img_max_min
+mosaicview(img, ramp, edge, edge_smoothed; nrow=2)
 
 # # References
+
 # Verbeek, P. W., Vrooman, H. A., & Van Vliet, L. J. (1988). [Low-level image processing by max-min filters](https://core.ac.uk/download/pdf/194053536.pdf). Signal Processing, 15(3), 249-258.
 
 ## save covers #src
 using ImageMagick #src
 mkpath("assets")  #src
-ImageMagick.save("assets/max_min.gif",cat(img,edge,edge_smoothed,map(clamp01nan,ramp);dims=3);fps =1) #src
+ImageMagick.save("assets/max_min.gif", cat(img, edge, edge_smoothed, map(clamp01nan, ramp);dims=3);fps=1) #src
--- a/docs/demos/filters/min_max_filter.jl
+++ b/docs/demos/filters/min_max_filter.jl
@@ -1,0 +1,57 @@
+# Topic : Max Min Filters
+# reference paper : https://core.ac.uk/download/pdf/194053536.pdf
+
+
+# In this tutorial we see how can we can effectively use max and min filter to distinguish 
+# between ramp("smooth") and ripple("texture") edges.We will using the mapwindow function of
+# ImageFiltering.jl package.
+
+# Note : If you are using the terminal use imshow(<Image var. name>) to view the image.
+
+
+using Images, ImageFiltering
+
+# We download an image from the internet and convert it into Grayscale image.
+# To work with your own image, use :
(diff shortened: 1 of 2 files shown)


docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,56 @@
+# Topic : Max Min Filters
+# reference paper : https://core.ac.uk/download/pdf/194053536.pdf
+
+
+# In this tutorial we see how can we can effectively use max and min filter to distinguish 
+# between ramp("smooth") and ripple("texture") edges of B&W image.We will using the mapwindow function of
```
timholy, line 6:
```suggestion
# between ramp("smooth") and ripple("texture") edges of B&W image. We will using the mapwindow function of
```

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,56 @@
...
+
+# No that we are done with the basic filtered images, we proceed to the next part
+# which is edge detection using these filters.
+
+# The trick is we define some thresholds.The thresholding can be dynamic for detecting
+# smooth parts or it can be texture based thresholding for detecting ripple edges.
+
+Dyt = (Max + Min)./ 2   # Dynamic Threshold(DYT)
+
+Tet = (Upp + Low)./ 2   # Texture Threshold(TET)
```
timholy, line 44: Where do these names come from?

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,56 @@
...
+# Note : If you are using the terminal use imshow(<Image var. name>) to view the image. 
+
+using Images, ImageFiltering
+
+# We download an image from the internet and convert it into Grayscale image.
+# To work with your own image, use :
+# img_path = "<FULL PATH TO THE IMAGE>"
+
+img_path = download("https://i.pinimg.com/originals/01/bc/8d/01bc8d82a3e2b4fa869f478479b97a3f.png")
+Img = Gray.(load(img_path));
```
timholy, line 18: Variables in julia should typically be lower-case. Matrices are an exception due to the legacy of math, otherwise it's better to follow the typical stylistic conventions:
- https://docs.julialang.org/en/v1/manual/style-guide/#Use-naming-conventions-consistent-with-Julia-base/
- https://github.com/invenia/BlueStyle

I'll only comment on this here, but it's relevant for all your variables.

timholy (commented):
There, I can proofread :laughing: . Still not having written a demo myself, I'd defer to @johnnychen94 about any DemoCard syntax.

docs/demos/filters/min_max_filter.jl
Dsantra92: The name comes from the paper referenced.

docs/demos/filters/min_max_filter.jl
Dsantra92: I am aware of the lowercase variable naming syntax, I was facing problem for the min and max which Julia assumes to be functions.

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,56 @@
...
+Min = mapwindow(minimum, Img, (5, 5))
+
+# Similarly for maximum
+Max = mapwindow(minimum, Img, (5, 5))
+
+# The max(min) filter
+Upp = mapwindow(maximum, Min, (9, 9))
+
+# The min(max) filter
+Low = mapwindow(minimum, Max, (9, 9))
```
johnnychen94, line 34: `#-` indicates a manual block breaking.

`##` means normal julia code comment, doing this we can squash images into one show in a row, which looks nicer to me. The following `det` things can be wrapped similarily.

```suggestion
#-

## Using the mapwindow, create an image min.
Min = mapwindow(minimum, Img, (5, 5))
## Similarly for maximum
Max = mapwindow(minimum, Img, (5, 5))
## The max(min) filter
Upp = mapwindow(maximum, Min, (9, 9))
## The min(max) filter
Low = mapwindow(minimum, Max, (9, 9))
mosaicview(Min, Max, Upp, Low; nrow=1)
```

image

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,56 @@
+# Topic : Max Min Filters
+# reference paper : https://core.ac.uk/download/pdf/194053536.pdf
```
johnnychen94, line 2: You can take a look at the existing demo on how to configure the meta data:

https://github.com/JuliaImages/ImageFiltering.jl/blob/5f4769ec0c8b0ec731aaa4c44ec6e6b33d1b0881/docs/demos/filters/median_filter.jl#L1-L6

I suggest we make a cover image for every demo we write; I don't like the default one (Although it's me that made the version...) Also refer to that example for how to generate a cover image (or gif).


image

docs/demos/filters/min_max_filter.jl
johnnychen94: How about `min_out`/`img_out_min`, `max_out`/`img_out_max`, etc?

docs/demos/filters/min_max_filter.jl
johnnychen94: 🤔 these names don't look good to me. Might need some explanation on what it means.

docs/demos/filters/min_max_filter.jl
johnnychen94, line 2: DemoCards make a relatively strong assumption that the first non-title paragraph is extracted as the description.

Just moving the "In this tutorial we see how can..." part ahead of these should work.

image

docs/demos/filters/min_max_filter.jl
johnnychen94, line 6: (Not important) word choice: `B&W image` => `grayscale image`

johnnychen94: `ramp("smooth")` and `ripple("texture")` looks really like a function call 😂 

How about "between the ramp/smooth and ripple/texture part of the grayscale image" ?

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,56 @@
...
+Tet = (Upp + Low)./ 2   # Texture Threshold(TET)
+
+# The Dynamic Gist(more focus on the the smooth edges)
+Dyg = Img - Dyt
+
+# The Texture Gist( isolates out the textures and/or noise)
+Teg = Img - Tet
+
+# Filtered Out edges
+Edge = Img - Low
```
johnnychen94, line 53: Actually, I haven't read through the details of that paper, is this some delicate design brought by that paper? If so, probably some more explanations can be useful to understand why we need to calculate these data.

docs/demos/filters/min_max_filter.jl
johnnychen94, line 2: URLs can easily vanish and becomes invalid. I suggest a formal citation format like what I did in #189 so that when the reader is interested, he always has a clue of what he can search for.

johnnychen94:
FYI, to preview the result locally, you can activate julia with `docs/` project, e.g.,

```bash
julia --project=docs/
```

and then run `include("docs/make.jl")` to generate the docs.

The generated HTML files can be found in `docs/build`, open it in your browser and you'll get a better feeling on the changes.

docs/demos/filters/min_max_filter.jl
timholy: Yes, there are many other names to choose from, so I'd rather follow the stylistic convention. In addition to the ones @johnnychen94  mentioned, `imgmin` and `imgmax` seem OK too.

Dsantra92:
Looks like I messed a lot.I am working on the fixes.

docs/demos/filters/min_max_filter.jl
Dsantra92: I was actually trying to keep a close correspondence to the paper :sweat_smile: .Anyway I had to remove those slashes too because it turns out demo cards can't render slashes.

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,72 @@
...
+
+# The average of min and max filters also gives us the smooth edges, but it also includes the noise in the image. 
+# So using average of `img_min` and `img_max` as a threshold  will yeild only texture.
+
+img_text_noise_threshold = (img_max_min + img_max_min)./ 2
+img_text_threshold = (img_max + img_min)./2
+mosaicview(img_text_noise_threshold, img_text_threshold; nrow = 1)
+#
+## The Dynamic Gist is obtained by subtracting the img_text_threshold from the original image.
+## The filtered image gives us the texture of the image.
```
johnnychen94, line 55:
```suggestion
## The Dynamic Gist gives us the texture of the image
```

No need to explain the obvious fact. More importantly, I still don't understand what "dynamic gist" means even googling it.

Similarily for "Texture Gist".

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,72 @@
...
+img_dynamic_gist = img - img_text_threshold
+## The Texture Gist is obtained by subtracting img_text_noise_threshold from the original image.
+## The filtered image gives us the texture along with the noise of the image.
+img_texture_gist = img - img_text_noise_threshold
+mosaicview(img_dynamic_gist, img_texture_gist; nrow = 1)
+#
+ramp = img_dynamic_gist - img_texture_gist
+## Filtered Out edges
+edge = img - img_min_max
+## Smoothed our version of edge
```
johnnychen94, line 65:
```suggestion
## Smoothed-out version of edge
```

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,72 @@
...
+edge_smoothed = img_max_min - img_min_max
+mosaicview(img, ramp, edge, edge_smoothed; nrow = 2)
+
```
johnnychen94, line 68: I notice that reference is removed in new commits, would you mind to add them back?

```suggestion

# # References

Verbeek, P. W., Vrooman, H. A., & Van Vliet, L. J. (1988). [Low-level image processing by max-min filters](https://core.ac.uk/download/pdf/194053536.pdf). Signal Processing, 15(3), 249-258.
```

This is probably confusing, `##` is a julia comment, and `# # Rerefences` is a markdown level-1 title; `# ` tells Literate.jl to filter this line into markdown format.

johnnychen94 (commented):
Looks much better!

A lot of language picks here. I'm not a native English speaker so I'm probably wrong:

* There is no space between the word and the next punctuation mark, and there is one space between the word and the previous punctuation mark. So in expressions "hello , world", "hello ,world", "hello,world", the space is not put correctly.
* Punctuation marks don't start a new line.
* there should be only one space between words.

For the same reason, most comments are written as `# this is a comment` instead of `#this is a comment`.

johnnychen94:
Oops sorry for the back-and-forth, I just realized that @timholy has implemented [a faster version of `maxmin` filter](https://github.com/JuliaImages/ImageFiltering.jl/blob/5f4769ec0c8b0ec731aaa4c44ec6e6b33d1b0881/src/mapwindow.jl#L323-L324) using the [Lemire filters](https://arxiv.org/abs/cs/0610046)

I'm not sure how this could be used efficiently, @timholy would you mind give some suggestions on this?

timholy:
`mapwindow(extrema, img, window)`.

docs/demos/filters/min_max_filter.jl
Dsantra92: The definition in the paper does not clearly explain what it is.It referred to some older paper which I could not find.

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,72 @@
...
+img_min_max = mapwindow(minimum, img_max, (3, 3))
+mosaicview(img_min, img_max, img_max_min, img_min_max; nrow = 1)
+
+# Now that we are done with the basic filtered images, we proceed to the next part
+# which is edge detection using these filters.
+
+# For edge detection we need define thresholds for our image.Threshold is an important tool 
+# for binarizing a grayscale image.The threshold value for a given pixel practically
+# decides if the pixel visible or not in the output image.Appyling a global threshold might underconsider the 
+# variation of colors/brightness within the image.So we consider an adaptive type theresholding.
```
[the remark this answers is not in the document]
Dsantra92: Yea got it. You sound a better teacher than the one I got in high school.

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,72 @@
+# ---
+# title: Max min filters
+# cover: assets/max_min.gif
+# author: Deeptendu Santra
+# date: 2020-10-08
+# ---
+
+# In this tutorial we see how can we can effectively use max and min filter to distinguish 
+# between smooth and texture edges  in grayscale images.
```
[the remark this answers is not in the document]
Dsantra92: I am not exactly sure are we using past tense here.

docs/demos/filters/min_max_filter.jl
Dsantra92: I think writing the Reference paper in markdown would be better than writing it in a cell.

docs/demos/filters/min_max_filter.jl
```diff
@@ -0,0 +1,72 @@
...
+using ImageCore,ImageShow, ImageFiltering
+using TestImages
+
+img = Gray.(testimage("lighthouse"))      #Original Image       
+
+# We can use a minimum function to check the minimum of GrayScale Values in the given matrix,or array
+# . For Example : 
+minimum([Gray(0.7),Gray(0.5),Gray(0.0)]) ##Should return Gray(0.0) i.e black.
+# 
+## Using the mapwindow, we create an image min.
```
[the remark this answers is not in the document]
Dsantra92: I first thought it that way but the line following it really messes the it up.

> Using the `mapwindow` function, we create an image of the local minimum.
 The mapwindow maps the function over a moving window of given size

We are using function in the next sentence for a different function which might just create a bit of confusion for the readers.Can we just do ?

> Using  mapwindow , we create an image of the local minimum.
The mapwindow maps the function over a moving window of given size.

Dsantra92:
I have made few changes as requested, hoping that it will make the cut.

docs/demos/filters/min_max_filter.jl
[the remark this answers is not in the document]
johnnychen94: I think `can` and `could` can be used interchangeably in many cases.

> http://unenlightenedenglish.com/2009/04/can-vs-could/#:~:text=Can%20is%20generally%20present%20tense,used%20to%20be%20more%20formal.
> The other use of could is to establish a level of doubt, tentativeness or politeness

johnnychen94:
It now looks awesome and I think it's ready to merge. Thanks for all the polishing work.

I've adjusted the filter size to make the result look nicer. If you are good with this, I can get it merged.

docs/demos/filters/min_max_filter.jl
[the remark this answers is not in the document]
johnnychen94: With the word `given`, I think your new commit looks clearer.

Dsantra92:
The new example looks good to me.

johnnychen94:
Thanks again for the work!

PR #191, opened 2020-10-08
Fix some ambiguities on nightly

timholy:
The new algorithm for detecting ambiguities picked up some real ones
that, surprisingly, were not detected previously.

Thanks to @vtjnash


timholy:
I don't think there's anything controversial here, so merging.

PR #193, opened 2020-10-11
also use the streaming max-min filter for maximum/minimum

johnnychen94:
I haven't read the Lemire paper yet so I don't know if there is a more efficient implementation for maximum/minimum, but now simply applying `first`/`last` on the `extrema` result gives about ~4x boost.

Hmmm, I plan to read that for my own interest in https://github.com/JuliaImages/Images.jl/issues/918.

cc @Dsantra92

### DIFF
--- a/docs/demos/filters/min_max_filter.jl
+++ b/docs/demos/filters/min_max_filter.jl
@@ -6,31 +6,47 @@
 # ---
 
 # In this tutorial we see how can we can effectively use max and min filter to distinguish
-# between smooth and texture edges  in grayscale images.
+# between smooth and texture edges in grayscale images. The example in this demo comes from [1].
 
 # We will use the [`mapwindow`](@ref) function in `ImageFiltering.jl` which provides a general
-# functionality to apply any function to the window around each pixel.
+# functionality to apply any function to the window around each pixel. 
+# [Custom median filters](@ref median_filter_example) is another usage example of `mapwindow`.
 
 using ImageCore, ImageShow, ImageFiltering
+using MappedArrays
 using TestImages
 
-img = Gray.(testimage("house");)      # Original Image
+img = Gray.(testimage("house")); # Original Image
 
-# We can use the `minimum` function to compute the minimum of the grayscale values in the given
-# matrix or array. For example: 
-minimum([Gray(0.7),Gray(0.5),Gray(0.0)]) # Should return Gray(0.0) i.e black.
-# 
-filter_size = (15, 15)
-## Using the `mapwindow` function, we create an image of the local minimum.
-## `mapwindow` maps the given function over a moving window of given size.
-img_min = mapwindow(minimum, img, filter_size)
-## Similarly for maximum
-img_max = mapwindow(maximum, img, filter_size)
-## The max(min) filter
-img_max_min = mapwindow(maximum, img_min, filter_size)
-## The min(max) filter
-img_min_max = mapwindow(minimum, img_max, filter_size)
+# We need four statistics for our demo, they are: the local minimum, maximum, min-max, max-min of
+# the input image. We can use `mapwindow` to get them.
+
+window_size = (15, 15)
+img_min = mapwindow(minimum, img, window_size)
+img_max = mapwindow(maximum, img, window_size)
+img_max_min = mapwindow(maximum, img_min, window_size)
+img_min_max = mapwindow(minimum, img_max, window_size)
 mosaicview(img_min, img_max, img_max_min, img_min_max; nrow=1)
+
+# When `f` is one of `maximum`, `minimum` and `maximum`, `mapwindow(f, img, window_size)` will use
+# the a streaming version Lemire max-min filter[2] to do the filtering work. This is more efficient
+# than a plain maximum implementation.
+
+# ```julia
+# using BenchmarkTools
+# f(x) = minimum(x) # do note that `f !== minimum`
+# @btime mapwindow(f, $img, window_size) # 47.508 ms (202831 allocations: 12.91 MiB)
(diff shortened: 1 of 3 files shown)


docs/demos/filters/min_max_filter.jl
```diff
@@ -6,32 +6,48 @@
...
+
+# ```julia
+# using BenchmarkTools
+# f(x) = minimum(x) # do note that `f !== minimum`
+# @btime mapwindow(f, $img, window_size) # 47.508 ms (202831 allocations: 12.91 MiB)
+# @btime mapwindow(minimum, $img, window_size) # 13.216 ms (58 allocations: 1.75 MiB)
+# ```
+
+# Also, in this example, since we need both the local minimum and maximum image. We could use
+# `extrema` to reduce the computation, and use `MappedArrays.mappedarray` to reduce allocation.
```
johnnychen94, line 43:
```suggestion
# `extrema` to reduce the computation, and use `MappedArrays.mappedarray` to reduce the allocation.
```

docs/demos/filters/min_max_filter.jl
```diff
@@ -6,32 +6,48 @@
...
+# This is more efficient than repeatedly doing `minimum` and `maximum`.
+
+img_extrema = mapwindow(extrema, img, window_size) # only compute once
```
johnnychen94, line 46: "only compute once" is ambiguous, since it is not clear that `minimum` result is dropped when calling  `mapwindow(maximum,  )`.

docs/demos/filters/min_max_filter.jl
```diff
@@ -6,32 +6,48 @@
...
+
+window_size = (15, 15)
+img_min = mapwindow(minimum, img, window_size)
+img_max = mapwindow(maximum, img, window_size)
+img_max_min = mapwindow(maximum, img_min, window_size)
+img_min_max = mapwindow(minimum, img_max, window_size)
 mosaicview(img_min, img_max, img_max_min, img_min_max; nrow=1)
 
+# When `f` is one of `maximum`, `minimum` and `maximum`, `mapwindow(f, img, window_size)` will use
+# the a streaming version Lemire max-min filter[2] to do the filtering work. This is more efficient
```
johnnychen94, line 32:
```suggestion
# a streaming version Lemire max-min filter[2] to do the filtering work. This is more efficient
```

timholy:
The speed improvement or worsening depends on the window size; I worry a 3x3 window might not show a speedup. The Lemire algorithm has a different order than naive `minimum`/`maximum`. The naive versions are `O(M^d*N)` (in `d` dimensions, where `M` is the window size and `N` the number of pixels) whereas my hazy memory suggests Lemire is `O(MlogM*N)`. For small windows the constant factor could be important.

The other issue is that if you want both the max & min, it's *clearly* better to call `mapwindow(extrema, ...)` once and then extract both results.

For these reasons I'd avoided implementing these shortcuts, in essence hoping to convince people to use `extrema` directly. But that may not be documented, and may not even be a good idea. So this PR may be a really good idea, but I thought you should know why we don't have this already.

johnnychen94:
It is still more efficient in 3*3 case, probably contributed by less memory allocation.

```julia
julia> @btime mapwindow(maximum, $img2d, (3,3));
  40.210 μs (54 allocations: 52.56 KiB)

julia> @btime mapwindow(f, $img2d, (3, 3));
  109.429 μs (2350 allocations: 141.25 KiB)
```

I'll need to read the paper first to see if there's a better way.

timholy:
Looks like the performance issue with `maximum` isn't memory allocation (though I haven't taken the time to figure out where that comes from, it's a bit surprising), it's that it is using a more careful `max`. The results for `img2d = rand(10,10); img2d[5,5] = NaN` seem to differ for a 3x3 window. :frowning:

johnnychen94:
Playing with https://github.com/sairus7/MaxMinFilters.jl and it turns out to be more performant than ImageFiltering's implementation. I guess this indicates that there's still room for optimization.

```julia
julia> img2d = randn(30,30);

julia> @btime mapwindow(extrema, $img2d, (3,1));
  21.749 μs (47 allocations: 44.91 KiB)

julia> @btime movmaxmin($img2d, 3);
  5.271 μs (4 allocations: 14.63 KiB)
```

MaxMinFilters does not apply to the multidimensional case, though; it simply treats it as a vector. I'll see what I can do to get things better here. A re-reimplementation, probably, in an independent package and get used by ImageFiltering.jl.

Just took the first read on the paper and it occurs that a maximum/minimum filter is possible if we only apply the wedge update on one side. So yeah, it's doable to for `mapwindow` to also support maximum/minimum.

Tweaking the implementation doesn't look like an easy job, so I'll revisit this when I finish my WNNM project. Probably months later, though...

timholy:
> A re-reimplementation, probably, in an independent package and get used by ImageFiltering.jl.

I'm unsure it should be a separate package...if anything, I'm inclined to go the other way (https://github.com/JuliaImages/Images.jl/issues/898). Well, I guess it could be a separate *package*, it's the separate *repo* that I find annoying.

timholy:
The main reason they differ in performance is *because* this implementation supports multiple dimensions. Half the time is due to this one comprehension: https://github.com/JuliaImages/ImageFiltering.jl/blob/0c963dc6bc4c465c95ef6a6bb3a49d030d3bbfce/src/mapwindow.jl#L375

But that's the foundation for multidimensional support, as it allows the filter to be separable. Almost all the rest is due internal methods of `DataStructures.CircularDeqeue`. I wonder if we'd get better performance out of an immutable variant?

Anyway, I don't think we should be so quick to abandon the current implementation. There might be a quick way to make the 1d case not need the comprehension, for example.

PR #205, opened 2021-02-08
Don't use FFT with non-finite input

yha:
Fixes `imfilter`ing vectors with ±Inf

### DIFF
--- a/test/nd.jl
+++ b/test/nd.jl
@@ -98,10 +98,14 @@ Base.zero(::Type{WrappedFloat}) = WrappedFloat(0.0)
         i = rand(eachindex(v))
         w = [j == i ? x : v[j] for j in eachindex(v)]
         kern = centered(ones(31))
+        # verify that non-finite values prevent use of FFT
+        @test ImageFiltering.filter_algorithm(v, v, kern) == ImageFiltering.FFT()
+        @test ImageFiltering.filter_algorithm(w, w, kern) == ImageFiltering.FIR()
         vf = imfilter(v, kern)
         wf = imfilter(w, kern)
-        @test all(isequal(x), wf[j] for j in (i-15:i+15) ∩ eachindex(v))
-        @test all(wf[j] ≈ vf[j] for j in eachindex(v) if abs(j-i) > 15)
+        around_i = [abs(i-j) <= 15 for j in eachindex(v)]
+        @test all(isequal(x), wf[around_i])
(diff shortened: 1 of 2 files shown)


test/nd.jl
```diff
@@ -92,9 +92,18 @@ Base.zero(::Type{WrappedFloat}) = WrappedFloat(0.0)
...
-    @test isnan.(imfilter([NaN;1:100],centered(ones(31)))) == ((0:100) .<= 15)
-
+    # Non-finite input
+    for x in (NaN, Inf, -Inf)
+        v = rand(100)
+        i = rand(eachindex(v))
+        w = [j == i ? x : v[j] for j in eachindex(v)]
+        kern = centered(ones(31))
+        vf = imfilter(v, kern)
+        wf = imfilter(w, kern)
```
timholy, line 102: To make sure that the spirit of this test is retained against future code changes (e.g., what if someone changes `L > 30` to `L > 45`?), we might want to explicitly test `ImageFiltering.filter_algorithm(w, w, kern) == FFT()`.

I acknowledge that I didn't do this carefully in the original test, so the lack isn't your fault.

test/nd.jl
```diff
@@ -92,9 +92,18 @@ Base.zero(::Type{WrappedFloat}) = WrappedFloat(0.0)
...
+        @test all(isequal(x), wf[j] for j in (i-15:i+15) ∩ eachindex(v))
+        @test all(wf[j] ≈ vf[j] for j in eachindex(v) if abs(j-i) > 15)
```
timholy, line 104: I worry this could be flaky: if `rand` generates some teeny numbers next to some not-teeny numbers, then roundoff differences between the two algorithms might cause this to fail. How about `wf[mask] ≈ vf[mask]`? That uses the norm of each vector as the scale of permitted deviations, rather than a separate scale for each numeric value, and might be more robust.

timholy (commented):
Great idea, thanks!

test/nd.jl
yha: Actually, I believe the original test here is mine too. Anyway, I added the algorithm check.

test/nd.jl
yha: Good point. Done.

timholy:
Thanks! Very nice change.

timholy:
https://github.com/JuliaRegistries/General/pull/29719

PR #214, opened 2021-05-22
use `OffsetArrays.centered`

johnnychen94:
requires:

- [x] ImageMetadata (https://github.com/JuliaImages/ImageMetadata.jl/pull/78)
~- [ ] AxisArrays (https://github.com/JuliaArrays/AxisArrays.jl/pull/196)~
- [x] https://github.com/JuliaImages/ImageAxes.jl/pull/57

Dependencies Requires and ImageMetadata are removed


src/ImageFiltering.jl
```diff
@@ -97,13 +99,6 @@ function __init__()
     #     @eval using DummyAF
     # end
     pop!(LOAD_PATH)
-    @require AxisArrays="39de3d68-74b9-583c-8d2d-e117c070f3a9" begin
-        centered(ax::AxisArrays.Axis{name}) where name = AxisArrays.Axis{name}(centered(ax.val))
-        centered(a::AxisArrays.AxisArray) = AxisArrays.AxisArray(centered(a.data), centered.(AxisArrays.axes(a)))
-    end
-    @require ImageMetadata="bc367c6b-8a6b-528e-b4bd-a4b897500b49" begin
-        centered(a::ImageMetadata.ImageMeta) = ImageMetadata.ImageMeta(centered(ImageMetadata.arraydata(a)), ImageMetadata.properties(a))
-    end
```
johnnychen94, line 106: There is a compatibility trick (prior) here: because ImageFiltering now depends on ImageCore v0.9, and ImageAxes and ImageMetadata versions that are compatible with ImageCore v0.9 have these `centered` methods defined, thus we can safely remove this.

One edge case is that users only load `ImageFiltering` and `AxisArrays` without loading `ImageAxes`, in which case `centered(::Axis)` will now error.

timholy: I am hoping to transition ImageAxes to a more-maintained named-axis package, perhaps we can address that there?

timholy (approved):

PR #218, opened 2021-06-22
WIP: Add template matching demo

### LINKED ISSUE #217: Template matching using mapwindow
Is there an example for template matching .
I think mapwindow could be used in combination with different functions.

After mapwindow a threshold to determine the matches could be calculated. I am just not sure how to combine matches that are close together.

feanor12:
Should fix #217 

I created an inital version, but I have some problems showing the resulting images.
The text is not final yet and also the code might need some improvement, but the example works.


docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,75 @@
...
+# Now we see small blobs at the locations which match our template and we can label the connected regions by label_components.
+# This will enumerate are connected regions and component_centroids can be used to get the centroid of each region. 
+# `component_centroids` also return the centroid for the backgroud region, which is at the first position and we will ommit it.
+
+centroids = component_centroids(label_components(threshold))[2:end]
+
+# To now see it it worked ciorrectly we can overlay the centroids with the original image using the Plots package.
+# As the images are stored 
+plot(Gray.(img))
+scatter!(reverse.(centroids),label="centroids",legend=:outertopleft)
```
johnnychen94, line 75: `#src` is to hide this line of code from the generated document while still running it; we use this to generate cover images. This cooperates with my suggestion in the `# cover: ` frontmatter.

```suggestion
scatter!(reverse.(centroids),label="centroids",legend=:outertopleft)

savefig("assets/template_matching.png") #src
```

docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,75 @@
+# ---
+# title: Template Matching
+# cover: assets/median.gif
```
johnnychen94, line 3: This cover can be generated during runtime:

```suggestion
# cover: assets/template_matching.png
```

docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,75 @@
...
+Gray.(template)
+#
+img = ones(100,100)
+img[(1:11).+20,(1:11).+50] .= template
+img[(1:11).+50,(1:11).+10] .= template
+img[(1:11).+70,(1:11).+80] .= template
+img[:,:] .*= rand(100,100) 
+Gray.(img)
+
+# Now that we have an image and a template we have to think have we measure the similarity between a
```
johnnychen94, line 40:
```suggestion
# Now that we have an image and a template, the next step is to define how we measure the similarity between a
```

docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,75 @@
...
+	
+# If the subsection is located at the border of the image the image has to be extended and in our case we will
+# fill all values outside the image with 1. As all of the square differences will be summed up per subsection 
+# the resulting sum can be bigger than 1. This will be a problem if we just convert it to an image to check  the values.
+# To rescale the values to be between 0 and 1 we can use imadjustintensity.
+
+rescaled_map = adjust_histogram(res, LinearStretching()) 
+
+#  To find the best locations we have to look for small values on the similarity map. This can be done by comparing
+# if the pixel are below a certain value. LÖets chose a value of 0.1.
```
johnnychen94, line 61:
```suggestion
# if the pixel is below a certain value. Let's chose a value of `0.1`.
```

docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,75 @@
...
+
+threshold = rescaled_map .< 0.1
+Gray.(threshold)
+
+# Now we see small blobs at the locations which match our template and we can label the connected regions by label_components.
```
johnnychen94, line 66:
```suggestion
# Now we see small blobs at the locations which match our template and we can label the connected regions by `label_components`.
```

docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,75 @@
...
+# This will enumerate are connected regions and component_centroids can be used to get the centroid of each region.
```
johnnychen94, line 67:
```suggestion
# This will enumerate are connected regions and `component_centroids` can be used to get the centroid of each region.
```

docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,75 @@
...
+
+#  To find the best locations we have to look for small values on the similarity map. This can be done by comparing
+# if the pixel are below a certain value. LÖets chose a value of 0.1.
+
+threshold = rescaled_map .< 0.1
+Gray.(threshold)
+
+# Now we see small blobs at the locations which match our template and we can label the connected regions by label_components.
+# This will enumerate are connected regions and component_centroids can be used to get the centroid of each region. 
+# `component_centroids` also return the centroid for the backgroud region, which is at the first position and we will ommit it.
```
johnnychen94, line 68:
```suggestion
# `component_centroids` also return the centroid for the background region, which is at the first position and we will omit it.
```

johnnychen94 (commented):
Thanks for adding this!

I find the random image is less attractive than a real image. Maybe we can use "cameraman" or "blobs" from TestImages?

Plots seem not very nice in generating the cover image, maybe we can use ImageDraw to draw the ROI and directly save it using `Images: save` from FileIO. BTW, `ImageCore.mosaicview` from MosaicVIews is also a good tool for visualization.

feanor12:
I am not sure why the math part does not work. maybe mathjax is missing.

docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,63 @@
...
+using ImageFiltering: mapwindow, Fill, imfilter, KernelFactors
+using ImageDistances: sqeuclidean
+using ImageContrastAdjustment: adjust_histogram, LinearStretching
+using TestImages
+using Plots: scatter!, plot
+
+# `ImageCore` enables the generation of images, `ImageFiltering` provides the [`mapwindow`](@ref) function and `ImageFeatures` provides functions to label segments of an image. To calculate the similarity `squeclidean` is used. This defines a distance between two images as:
+#
+# ```math
+# distance = \sum_i^[number of pixels} (template_i - subsection_i)^2
```
johnnychen94, line 24: maybe? 
```suggestion
# distance = \sum_i^{number of pixels} (template_i - subsection_i)^2
```

docs/demos/search/template_matching.jl
```diff
@@ -0,0 +1,63 @@
...
+# ```
+#
+# `ImageContrastAdjustment` provides functions to adjust the histogram, which is useful when an image contains values bigger than 1 or smaller than 0.  The `Testimages` package will provide our image and plot is used to overlay a scatter plot onto the image.
+
+# To start we first load our image.
+img = testimage("lena_gray_512")
```
johnnychen94, line 30: For some reason, we don't recommend to use lena in new codes (https://github.com/JuliaImages/TestImages.jl/issues/72).

feanor12: Ok, I'll try to use another one.

feanor12:
I hope moonsurface is ok to use. :)

johnnychen94:
nice! Thank you very much for adding new demos!

PR #221, opened 2021-06-28
update error check msg

johnnychen94:


johnnychen94:
DemoCards v0.4.3 is required to let docs CI pass and it's currently not available for the Pkg server due to its sync latency. I'll merge this PR tomorrow to give PkgServer some time to update the registry it serves.

PR #223, opened 2021-09-01
Move extremization, blob_LoG from Images

timholy:
I think (?) the best way to do this is to make a minor bump. So we should make sure we move everything at once. Am I missing things? What about `imROF`?

xref https://github.com/JuliaImages/ImageSegmentation.jl/pull/72

### DIFF
--- a/src/ImageFiltering.jl
+++ b/src/ImageFiltering.jl
@@ -18,7 +18,9 @@ export Kernel, KernelFactors,
     imfilter, imfilter!,
     mapwindow, mapwindow!,
     imgradients, padarray, centered, kernelfactors, reflect,
-    freqkernel, spacekernel
+    freqkernel, spacekernel,
+    findlocalminima, findlocalmaxima,
+    blob_LoG, BlobLoG
 
 FixedColorant{T<:Normed} = Colorant{T}
 StaticOffsetArray{T,N,A<:StaticArray} = OffsetArray{T,N,A}
@@ -56,6 +58,7 @@ using .Algorithm: Alg, FFT, FIR, FIRTiled, IIR, Mixed
 Alg(r::AbstractResource{A}) where {A<:Alg} = r.settings
 
 include("utils.jl")
+include("compat.jl")
 include("kernelfactors.jl")
 using .KernelFactors: TriggsSdika, IIRFilter, ReshapedOneD, iterdims, kernelfactors
 
@@ -85,6 +88,7 @@ include("specialty.jl")
 
 include("mapwindow.jl")
 using .MapWindow
+include("extrema.jl")
 
 function __init__()
     # See ComputationalResources README for explanation
--- a/src/extrema.jl
+++ b/src/extrema.jl
@@ -1,0 +1,117 @@
+"""
+BlobLoG stores information about the location of peaks as discovered by `blob_LoG`.
+It has fields:
+
+- location: the location of a peak in the filtered image (a CartesianIndex)
+- σ: the value of σ which lead to the largest `-LoG`-filtered amplitude at this location
+- amplitude: the value of the `-LoG(σ)`-filtered image at the peak
+
+Note that the radius is equal to σ√2.
+
+See also: [`blob_LoG`](@ref).
+"""
+struct BlobLoG{T,S,N}
+    location::CartesianIndex{N}
+    σ::S
+    amplitude::T
+end
+
+"""
+    blob_LoG(img, σscales; edges=(true, false, ...), σshape=(1, ...)) -> Vector{BlobLoG}
+
+Find "blobs" in an N-D image using the negative Lapacian of Gaussians
+with the specifed vector or tuple of σ values. The algorithm searches for places
+where the filtered image (for a particular σ) is at a peak compared to all
+spatially- and σ-adjacent voxels, where σ is `σscales[i] * σshape` for some i.
+By default, `σshape` is an ntuple of 1s.
+
+The optional `edges` argument controls whether peaks on the edges are
+included. `edges` can be `true` or `false`, or a N+1-tuple in which
+the first entry controls whether edge-σ values are eligible to serve
+as peaks, and the remaining N entries control each of the N dimensions
+of `img`.
+
+# Citation:
+
+Lindeberg T (1998), "Feature Detection with Automatic Scale Selection",
+International Journal of Computer Vision, 30(2), 79–116.
+
+See also: [`BlobLoG`](@ref).
+"""
+function blob_LoG(img::AbstractArray{T,N}, σscales::Union{AbstractVector,Tuple};
+                  edges::Union{Bool,Tuple{Vararg{Bool}}}=(true, ntuple(d->false, Val(N))...), σshape=ntuple(d->1, Val(N))) where {T,N}
+    if edges isa Bool
+        edges = (edges, ntuple(d->edges,Val(N))...)
+    end
+    sigmas = sort(σscales)
+    img_LoG = Array{Float64}(undef, length(sigmas), size(img)...)
+    colons = ntuple(d->Colon(), Val(N))
+    @inbounds for isigma in eachindex(sigmas)
+        img_LoG[isigma,colons...] = (-sigmas[isigma]) * imfilter(img, Kernel.LoG(ntuple(i->sigmas[isigma]*σshape[i],Val(N))))
+    end
+    maxima = findlocalmaxima(img_LoG; dims=1:ndims(img_LoG), edges=edges)
+    [BlobLoG(CartesianIndex(tail(x.I)), sigmas[x[1]], img_LoG[x]) for x in maxima]
+end
+
+
+"""
+    findlocalmaxima(img; dims=coords_spatial(img), edges=true) -> Vector{CartesianIndex}
+
+Returns the coordinates of elements whose value is larger than all of
+their immediate neighbors.  `dims` is a list of dimensions to
+consider.  `edges` is a boolean specifying whether to include the
+first and last elements of each dimension, or a tuple-of-Bool
+specifying edge behavior for each dimension separately.
+"""
+findlocalmaxima(img::AbstractArray; dims=coords_spatial(img), edges=true) =
+        findlocalextrema(img, dims, edges, Base.Order.Forward)
+
+"""
+    findlocalminima(img; dims=coords_spatial(img), edges=true) -> Vector{CartesianIndex}
+
+Like [`findlocalmaxima`](@ref), but returns the coordinates of the smallest elements.
+"""
+findlocalminima(img::AbstractArray; dims=coords_spatial(img), edges=true) =
+    findlocalextrema(img, dims, edges, Base.Order.Reverse)
+
+
+findlocalextrema(img::AbstractArray{T,N}, dims, edges::Bool, order) where {T,N} = findlocalextrema(img, dims, ntuple(d->edges,Val(N)), order)
+
+function findlocalextrema(img::AbstractArray{T,N}, dims, edges::NTuple{N,Bool}, order::Base.Order.Ordering) where {T<:Union{Gray,Number},N}
+    dims ⊆ 1:ndims(img) || throw(ArgumentError("invalid dims"))
+    extrema = Array{CartesianIndex{N}}(undef, 0)
+    Iedge = CartesianIndex(map(!, edges))
+    R0 = CartesianIndices(img)
+    R = clippedinds(R0, Iedge)
+    I1 = _oneunit(first(R0))
+    Rinterior = clippedinds(R0, I1)
+    iregion = CartesianIndex(ntuple(d->d ∈ dims, Val(N)))
+    Rregion = CartesianIndices(map((f,l)->f:l,(-iregion).I, iregion.I))
+    z = zero(iregion)
+    for i in R
+        isextrema = true
+        img_i = img[i]
+        if i ∈ Rinterior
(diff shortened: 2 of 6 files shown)


timholy:
Is the doc failure (segfault :scream:) any reason to hold off on merging this?

johnnychen94:
The documentation segfault might be related to Plots, see also https://github.com/JuliaImages/juliaimages.github.io/pull/210. I haven't had the time to investigate it so the best strategy is to temporarily remove it.

src/extrema.jl
```diff
@@ -0,0 +1,117 @@
...
+of `img`.
+
+# Citation:
+
+Lindeberg T (1998), "Feature Detection with Automatic Scale Selection",
+International Journal of Computer Vision, 30(2), 79–116.
+
+See also: [`BlobLoG`](@ref).
+"""
+function blob_LoG(img::AbstractArray{T,N}, σscales::Union{AbstractVector,Tuple};
```
johnnychen94, line 41: Maybe deprecate `AbstractVector` input for `σscales`?

src/extrema.jl
```diff
@@ -0,0 +1,117 @@
...
+                  edges::Union{Bool,Tuple{Vararg{Bool}}}=(true, ntuple(d->false, Val(N))...), σshape=ntuple(d->1, Val(N))) where {T,N}
+    if edges isa Bool
+        edges = (edges, ntuple(d->edges,Val(N))...)
+    end
+    sigmas = sort(σscales)
+    img_LoG = Array{Float64}(undef, length(sigmas), size(img)...)
+    colons = ntuple(d->Colon(), Val(N))
+    @inbounds for isigma in eachindex(sigmas)
+        img_LoG[isigma,colons...] = (-sigmas[isigma]) * imfilter(img, Kernel.LoG(ntuple(i->sigmas[isigma]*σshape[i],Val(N))))
```
johnnychen94, line 50: It looks like we can use `imfilter!` here.

src/extrema.jl
```diff
@@ -0,0 +1,117 @@
...
+
+"""
+    findlocalmaxima(img; dims=coords_spatial(img), edges=true) -> Vector{CartesianIndex}
+
+Returns the coordinates of elements whose value is larger than all of
+their immediate neighbors.  `dims` is a list of dimensions to
+consider.  `edges` is a boolean specifying whether to include the
+first and last elements of each dimension, or a tuple-of-Bool
+specifying edge behavior for each dimension separately.
+"""
```
johnnychen94, line 65: Maybe add some small examples to this and `findlocalminima`?

src/extrema.jl
```diff
@@ -0,0 +1,117 @@
...
+    Rregion = CartesianIndices(map((f,l)->f:l,(-iregion).I, iregion.I))
+    z = zero(iregion)
+    for i in R
+        isextrema = true
+        img_i = img[i]
+        if i ∈ Rinterior
+            # If i is in the interior, we don't have to worry about i+j being out-of-bounds
+            for j in Rregion
+                j == z && continue
+                if !Base.Order.lt(order, img[i+j], img_i)
```
johnnychen94, line 98: I was actually thinking if it's better to expose this as argument `f` and make

```julia
findlocalextrema(f, img; [dims,] [edges])
```

johnnychen94:
> Am I missing things? What about imROF?

I guess `gaussian_pyramid` (I've commented on https://github.com/JuliaImages/Images.jl/pull/971#issuecomment-910271594)

src/extrema.jl
johnnychen94: And furthermore, maybe it's better to have it work similar to `mapwindow`:

```julia
findall_window(f, img, window; border, indices)
```

src/extrema.jl
timholy: That would destroy the original image, and we need to filter the original with a range of different `σ`. Or are you proposing that we first replicate and then filter in-place?

johnnychen94: I meant this:

```julia
    LoG_slice = @view img_LoG[isigma, colons...]
    imfilter!(LoG_slice, img, ...)
    LoG_slice .*= -sigmas[isigma]
```

src/extrema.jl
timholy: I was in initial agreement but actually `σscales` is not guaranteed to be small. It doesn't correspond to, say, array dimensions, instead it will be the size of the first dimension of `img_LoG`. There's no reason that couldn't be 20 or higher, at which point using tuples doesn't make a lot of sense.

src/extrema.jl
timholy: That's pretty interesting. From a pragmatic standpoint one would probably want to smooth with a separable filter and then use a 3x3 window rather than an nxn window (the latter is O(n^2), the former is O(n)). But I do see the point.

src/ImageFiltering.jl
```diff
@@ -18,7 +18,9 @@ export Kernel, KernelFactors,
     imfilter, imfilter!,
     mapwindow, mapwindow!,
     imgradients, padarray, centered, kernelfactors, reflect,
-    freqkernel, spacekernel
+    freqkernel, spacekernel,
+    findlocalminima, findlocalmaxima,
+    blob_LoG, BlobLoG
```
johnnychen94, line 23: Do we really need to export the struct `BlobLoG`?

src/extrema.jl
```diff
@@ -0,0 +1,166 @@
...
+function findlocalextrema(f::F, img::AbstractArray{T,N}, window::Dims{N}, edges::NTuple{N,Bool}) where {F,T<:Union{Gray,Number},N}
+    extrema = Vector{CartesianIndex{N}}(undef, 0)
+    Iedge = CartesianIndex(map(!, edges))
+    R0 = CartesianIndices(img)
+    R = clippedinds(R0, Iedge)
+    halfwindow = CartesianIndex(map(x -> x >> 1, window))
+    Rinterior = clippedinds(R0, halfwindow)
+    Rwindow = _colon(-halfwindow, halfwindow)
+    z = zero(halfwindow)
+    for i in R
```
johnnychen94, line 140: It looks valid for me to skip the default bounds checking here:

```suggestion
    @inbounds for i in R
```

johnnychen94 (approved):
I'm thinking of an even generic window version of `findall` that iterates over all the interior points of `img` and return all `p` that satisfies `f(window_p) == true`. Just like `mapwindow` is the window version of `map`, I'm thinking it as the window version of `findall`.

Since we need this by https://github.com/JuliaImages/ImageSegmentation.jl/pull/72, maybe we can leave this as future work if that sounds too much coding work.

timholy:
> I'm thinking of an even generic window version of findall that iterates over all the interior points of img and return all p that satisfies f(window_p) == true. Just like mapwindow is the window version of map, I'm thinking it as the window version of findall...maybe we can leave findall_window as future work if that sounds too much coding work.

An excellent design concept and a worthy goal. No time like the present...

src/ImageFiltering.jl
timholy: Together with my kwarg method, it makes the output cut/pasteable ("round-trippable"), but maybe that's not important.

johnnychen94:
> No time like the present...

Yeah... I'm also running short of time..

Merge when you ready.

timholy:
Sorry, I meant "there is no better time than the present." So I'll try it now. (Yes, I am busy too, but your suggestion is too good to leave for later.)

johnnychen94:
> Should f be f(V, base_index) where V is the view of img?

My understanding on `findall` is that we are comparing the values and not the index. If we don't consider border condition, it might be as simple as:

```julia
results = CartesianIndex{N}[]
for (window_ax, i) in zip(TileIterator(axes(X), window)), CartesianIndices(X))
    window = @view X[window_ax...]
    f(window) && push!(results, i)
end
```

Well. Some efforts are needed to achieve the current performance of `findlocalextrema`, however. I did try this before I commented https://github.com/JuliaImages/ImageFiltering.jl/pull/223#pullrequestreview-744624172 but it seems 2x slower.

timholy:
Currently we exclude windows where `A[j] == A[i]` when `i != j`, and several of our tests fail if we change that.

timholy:
OK, see if you like this API.

Performance:
```julia
julia> A = rand(100, 101);

julia> @btime findlocalmaxima($A);
  132.642 μs (11 allocations: 64.47 KiB)

julia> @btime ImageFiltering.findmax_window($A, $(ImageFiltering.default_window(A)));
  182.121 μs (11 allocations: 64.47 KiB)
```

So not too shabby, but probably enough of a difference that we still want `findlocalmaxima`.

johnnychen94:
Will `f(img[centerpoint], img[otherpoint])` too strict assumption here? I think `f(Rview) == true` should be the most generic case. Recall my slow `mapwindow` example, https://github.com/JuliaImages/ImageSegmentation.jl/pull/72#issuecomment-908251756 I feel it's more natural to write the do-block here.

Of course, the performance is really an issue.

timholy:
My docs were lying a bit, I combined two things I shouldn't have combined. This is more accurate (the code itself hasn't changed).

timholy:
Did you see this (edited) comment:

> findall_window(>=, zeros(m, n), window) would return all points in the array!

That's the core problem with your implementation in https://github.com/JuliaImages/ImageSegmentation.jl/pull/72#issuecomment-908251756.

johnnychen94:
> OK, see if you like this API.

Right, I think I'm thinking of the same implementation as your current version. I'll try more if there are better ideas to optimize the performance and then will comment.

If we allow paddings like `mapwindow` then we don't need to pass extra `i` to `f(OffsetArray(view(img, Rview), Rview.indices), i)`; we can simply pass the local coordinates `f(Rview)`.

timholy:
Not sure I understand this:

> we can simply pass the local coordinates f(Rview)

That only passes the indices, and leaves the array itself implicit. That's pretty different from how `mapwindow` works. Is that really what you mean? If so, then you don't even need to pass `img` to `mapwindow` and `findall_window`, and that's different from `map` and `findall`.

src/findall_window.jl
```diff
@@ -0,0 +1,51 @@
...
+"""
+    all_window(f, V, basepoint, excludepoint=nothing)
+
+Returns `true` if `f(img[centerpoint], img[otherpoint])` for all indices `otherpoint`.
+Optionally exclude a single point (e.g., `basepoint` itself) by setting `excludepoint`.
+"""
+@inline function all_window(f::F, V::AbstractArray, basepoint::CartesianIndex, excludepoint::Union{Nothing,CartesianIndex}=nothing) where F
+    @inbounds ref = V[basepoint]
+    @inbounds for (i, v) in pairs(V)
+        i == excludepoint || f(ref, v) || return false
```
johnnychen94, line 31: I'm expecting `f(ref, v)` is scalar operation and is type stable, but if I change to

```julia
i == excludepoint || ref == v || return false
```

I get 

```julia
julia> @btime ImageFiltering.findmax_window($A, $(ImageFiltering.default_window(A)));
  74.347 μs (1 allocation: 80 bytes)
```

Surprisingly less allocations, do you have any ideas?

timholy: Presumably the output array is empty? So it's probably just `push!` reallocating the output buffer.

johnnychen94: With

```julia
function findall_window(f::F, img::AbstractArray{T,N}, window::Dims{N}; allow_edges::NTuple{N,Bool}=ntuple(_->true, N)) where {F,T<:Union{Gray,Number},N}
    buffer_size = min(512, length(img)) # a heuristic threshold that to work around the overhead of `push!`
    basepoints = Vector{CartesianIndex{N}}(undef, buffer_size)
    Iedge = CartesianIndex(map(!, allow_edges))
    R0 = CartesianIndices(img)
    R = clippedinds(R0, Iedge)
    halfwindow = CartesianIndex(map(x -> x >> 1, window))
    i_positive = 0
    @inbounds for i in R
        Rview = _colon(i-halfwindow, i+halfwindow) ∩ R0
        if f(OffsetArray(view(img, Rview), Rview.indices), i)
            i_positive += 1
            if length(basepoints) < i_positive
                resize!(basepoints, min(2length(basepoints), length(img)))
            end
            basepoints[i_positive] = i
        end
    end
    return resize!(basepoints, i_positive)
end
```

I'm getting closer:

```julia
julia> @btime ImageFiltering.findmax_window($A, $(ImageFiltering.default_window(A)));
  146.643 μs (4 allocations: 60.31 KiB) # after
  167.950 μs (11 allocations: 64.47 KiB) # before
```

timholy: `sizehint!` does the same thing and is much simpler.

timholy: And these optimizations could be applied to `findlocalextrema`, widening the gap once again. (The same comment applies to my own `i == excludepoint || ...` change.)

I think the key question is any overhead from creating the wrapper, which `findlocalextrema` doesn't do.

timholy:
Should we split out the `findall_window` into a separate PR? Or settle on it here?

johnnychen94:
Since the API is settled we can have both here and then try to improve the performance in future PRs.

src/findall_window.jl
```diff
@@ -0,0 +1,51 @@
+"""
+    findall_window(f, img::AbstractArray{T,N}, window::Dims{N}; allow_edges::(true...,))
```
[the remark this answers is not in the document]
timholy: It depends on what you mean by "correctly." The sense here is, "will it return a solution on the edge of the image?" In that sense it works fine (I think). What sense did you mean?

PR #225, opened 2021-09-03
Add findall_window

### LINKED ISSUE #224: introduce a `findall` version of `mapwindow`
Originally posted in https://github.com/JuliaImages/ImageFiltering.jl/pull/223#pullrequestreview-744624172

We have `mapwindow` that applies function `f` on every local patch of an image, which can be seen as the image version of `map`. We don't have the window version of `findall` that applies `f` to each local patch, and return coordinates such that `f(window_of_p) == true`.

Why we need this? We can currently do this by returning a dense Bool matrix from `mapwindow` and then call `findall`, but this would usually be unnecessary and not as performant as the manually written for-loops. For example, the benchmark in https://github.com/JuliaImages/ImageSegmentation.jl/pull/72#issuecomment-910135541

timholy:
Continued from #223

Closes #224

### DIFF
--- a/src/ImageFiltering.jl
+++ b/src/ImageFiltering.jl
@@ -16,7 +16,7 @@ export Kernel, KernelFactors,
     BorderArray,
     Algorithm,
     imfilter, imfilter!,
-    mapwindow, mapwindow!,
+    mapwindow, mapwindow!, findall_window,
     imgradients, padarray, centered, kernelfactors, reflect,
     freqkernel, spacekernel,
     findlocalminima, findlocalmaxima,
@@ -89,6 +89,7 @@ include("specialty.jl")
 include("mapwindow.jl")
 using .MapWindow
 include("extrema.jl")
+include("findall_window.jl")
 
 function __init__()
     # See ComputationalResources README for explanation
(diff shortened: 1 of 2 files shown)


johnnychen94:
I haven't checked it very carefully, but maybe this is related to the performance gap to `findlocalmaxima`

```julia
function sum_window_mean_ver1(X)
    out = zero(float(eltype(X)))
    R = CartesianIndices(X)
    Δ = oneunit(eltype(R))
    @inbounds @simd for p in R
        window = max(p-Δ, first(R)):min(p+Δ, last(R))
        out += mean(view(X, window))
    end
    return out
end

function sum_window_mean_ver2(X)
    out = zero(float(eltype(X)))
    R = CartesianIndices(X)
    Δ = oneunit(eltype(R))
    @inbounds @simd for p in R
        window = max(p-Δ, first(R)):min(p+Δ, last(R))
        mean_val = zero(eltype(out))
        for q in window
            mean_val += X[q]
        end
        out += mean_val/length(window)
    end
    return out
end
```

```julia
julia> @btime sum_window_mean_ver1($X);
  164.748 μs (0 allocations: 0 bytes)
julia> @btime sum_window_mean_ver2($X)
  64.137 μs (0 allocations: 0 bytes)
```

timholy:
Did you try creating your own `mean` function in case it's something about the one in `Statistics`?

PR #226, opened 2021-09-04
Add `box` and `laplacian2d`

timholy:
These come from Images.jl and are part of moving most/all code
out to make it a meta-package.

As annotated in https://github.com/JuliaImages/Images.jl/pull/971, these may be the last of the moves here (`gaussian_pyramid` does not seem likely to move). Consequently, once this merges we could release a new version, unblocking https://github.com/JuliaImages/ImageSegmentation.jl/pull/72, or we could wait for #214 and/or #225.

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -37,6 +37,14 @@ function product2d(kf)
 end
 
 """
+    kern = box(m, n)
+    kern = box((m, n, ...))
+
+Return a box kernel computing a moving average. `m, n, ...` specify the size of the kernel, which is centered around zero.
+"""
+box(sz::Dims) = broadcast(*, KernelFactors.box(sz)...)
+
+"""
 ```julia
     diff1, diff2 = sobel()
 ```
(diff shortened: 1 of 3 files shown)


src/kernel.jl
```diff
@@ -380,6 +388,40 @@ function Base.convert(::Type{AbstractArray}, L::Laplacian{N}) where N
...
+
+julia> Kernel.laplacian2d(0.5)    # equal weight for face-pixels and corner-pixels.
+3×3 OffsetArray(::Matrix{Float64}, -1:1, -1:1) with eltype Float64 with indices -1:1×-1:1:
+ 0.333333   0.333333  0.333333
+ 0.333333  -2.66667   0.333333
+ 0.333333   0.333333  0.333333
+```
+"""
+function laplacian2d(alpha::Number=0)
+    lc = alpha/(1 + alpha)
```
johnnychen94, line 420: Should `alpha > -1` here?

johnnychen94 (commented):
> As annotated in JuliaImages/Images.jl#971, these may be the last of the moves here (gaussian_pyramid does not seem likely to move)

👍

src/kernel.jl
timholy: Certainly `alpha = -1` is a bad choice. But I can't say that
```julia
julia> Kernel.laplacian2d(-2)
3×3 OffsetArray(::Matrix{Float64}, -1:1, -1:1) with eltype Float64 with indices -1:1×-1:1:
  2.0  -3.0   2.0
 -3.0   4.0  -3.0
  2.0  -3.0   2.0
```
would never be useful to anyone. And it has some laplacian-like properties, such as a sum of 0, symmetry, and satisfaction of Green's-like identities. 

I wish this had come with a reference, so we could check a citation, but believe it or not I think this predates my own 2012 involvement with Julia!

src/kernel.jl
```diff
@@ -36,6 +36,14 @@ function product2d(kf)
     k1[1].*k1[2], k2[1].*k2[2]
 end
 
+"""
+    kern = box(m, n)
+    kern = box((m, n, ...))
+
+Return a box kernel computing a moving average. `m, n, ...` specify the size of the kernel, which is centered around zero.
+"""
+box(sz::Dims) = broadcast(*, KernelFactors.box(sz)...)
```
johnnychen94, line 45:
```suggestion
box(sz::Dims) = broadcast(*, KernelFactors.box(sz)...)
box(sz::Int...) = box(sz)
```

test/specialty.jl
```diff
@@ -150,6 +150,19 @@ using ImageFiltering: IdentityUnitRange
...
+                                                   1/3 -8/3  1/3;
+                                                   1/3  1/3  1/3])
+    end
+
+    @testset "box" begin
+        @test KernelFactors.box(3) == (centered([1/3, 1/3, 1/3]),)
+        @test KernelFactors.box(3, 5) == (centered(reshape([1/3, 1/3, 1/3], 3, 1)), centered([1/5, 1/5, 1/5, 1/5, 1/5]'))
+        @test KernelFactors.box((3,5,7)) == KernelFactors.box(3, 5, 7)
+        @test Kernel.box((3,)) == centered([1/3, 1/3, 1/3])
+        @test Kernel.box((3, 5)) == centered([1/3, 1/3, 1/3] .* [1/5, 1/5, 1/5, 1/5, 1/5]')
```
johnnychen94, line 163:
```suggestion
        @test Kernel.box((3, 5)) == centered([1/3, 1/3, 1/3] .* [1/5, 1/5, 1/5, 1/5, 1/5]')
        @test Kernel.box(3, 5) == Kernel.box((3, 5))
```

johnnychen94 (approved):

src/kernel.jl
timholy: Sadly we can't really do that. Others are specialized to assume 2d with a single input :frowning_face: .

johnnychen94: Hmmm, then I guess we need to remove the `kern = box(m, n)` signature from the docstring?

timholy: Oh, right, thanks!

PR #227, opened 2021-09-05
Fix Plots segfaults when generating demos

johnnychen94:
I have no idea if this works on CI environments so let's try and see.


johnnychen94:
:tada:

timholy:
Thanks so much!

PR #230, opened 2021-09-26
Nice display for ReshapedOneD

timholy:
While `ReshapedOneD` is not itself an AbstractArray (due to needing to
handle IIR types too), it often contains one. In such cases, there are
obvious mechanisms to improve the display.

With this PR, we get:
```julia
julia> kern1, kern2 = kernelfactors((centered([1, 2, 1]), [-1, 0, 1]));

julia> kern1
Reshaped 1d stencil with axes (-1:1, 0:0):
 1
 2
 1

julia> kern2
Reshaped 1d stencil with axes (0:0, 1:3):
 -1  0  1
```


timholy:
CC @chriselrod xref https://github.com/JuliaSIMD/LoopVectorization.jl/issues/342#issuecomment-927299004

PR #231, opened 2021-09-27
Reduce unnecessary allocations in `filtfft`

BioTurboNick:
Several intermediate steps in `filtfft` are allocating when they don't need to be.

Results of this change on one of my operations:

Before: 86.106 ms (200 allocations: 140.55 MiB)
After: 72.876 ms (196 allocations: 113.27 MiB)

19% improvement in time, 24% improvement in allocation size.


johnnychen94 (approved):

timholy:
Thanks @BioTurboNick!

PR #234, opened 2021-10-30
docs: fix build warnings and run doctest as unit test

johnnychen94:
Will merge once test passes and then tag a new release


timholy:
Thanks for fixing this!

PR #235, opened 2021-11-01
rework gabor filter and add log gabor filter

johnnychen94:
The previous `Kernel.gabor` function has three drawbacks:

- it eagerly allocates memory, and thus makes optimization quite difficult
- it returns `Tuple{Matrix, Matrix}` as real and imaginary part; we should just return `Matrix{Complex{Float64}}` instead
- it doesn't have good documentation or examples... It takes quite a while to understand it..

benchmark on gabor filter:

```julia
using ImageFiltering

# new lazy Gabor array
kern = Kernel.Gabor((256, 256); orientation=0, phase_offset=0, wavelength=30, bandwidth=2, aspect_ratio=0.5)
@btime collect($kern); # 1.157 ms (2 allocations: 1.00 MiB)

# old gabor function
σ, θ, λ, γ, ψ = 10, 0, 30, 0.5, 0
@btime Kernel.gabor(256,256,σ,θ,λ,γ,ψ); # 2.911 ms (51 allocations: 12.60 MiB)
```

TODO:

- [ ] add tests
- [ ] revisit the implementation details
- [ ] add doc for log gabor

The implementation of the log gabor filter heavily follows https://www.peterkovesi.com/matlabfns/PhaseCongruency/Docs/convexpl.html so I'm wondering if @peterkovesi could help review this.


johnnychen94:
I'm marking this as ready. I'm going to add phase congruency filter next, and then finally the FSIM quality index https://github.com/JuliaImages/ImageQualityIndexes.jl/issues/25

ImagePhaseCongruency.jl is really a good reference but it has a strong MATLAB legacy, I believe we can get things faster and better composability by properly resolving and designing the function.

johnnychen94:
Benchmark shows that F32 on LogGabor is somewhat slow:

```julia
    BenchmarkGroup:
        2-element BenchmarkTools.BenchmarkGroup:
          tags: []
          "Gabor" => 6-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "Gabor_ComplexF32_19×19" => Trial(5.625 μs)
                  "Gabor_ComplexF32_256×256" => Trial(994.030 μs)
                  "Gabor_ComplexF32_512×512" => Trial(3.857 ms)
                  "Gabor_ComplexF64_19×19" => Trial(7.870 μs)
                  "Gabor_ComplexF64_256×256" => Trial(1.440 ms)
                  "Gabor_ComplexF64_512×512" => Trial(5.553 ms)
          "LogGabor" => 6-element BenchmarkTools.BenchmarkGroup:
                  tags: []
                  "LogGabor_ComplexF64_19×19" => Trial(13.903 μs)
                  "LogGabor_ComplexF64_256×256" => Trial(2.422 ms)
                  "LogGabor_ComplexF64_512×512" => Trial(9.884 ms)
                  "LogGabor_ComplexF32_19×19" => Trial(68.269 μs)
                  "LogGabor_ComplexF32_256×256" => Trial(11.896 ms)
                  "LogGabor_ComplexF32_512×512" => Trial(51.113 ms)
```

---

Writing benchmark isn't something pleasant, but when you get the report, oh that's a surprise!

peterkovesi:
I am very pleased and impressed that this work is being done. Please help yourself to any code you think is useful in ImagePhaseCongruency I am sure much of it can be considerably improved and made more interoperable with the rest of the Images ecosystem.

Another function from ImagePhaseCongruency that I think is deserving of being included in Images is `perfft2()`.  It is very useful when one wants to perform filtering in the frequency domain with minimal artifacts introduced from the boundary discontinuities.  See <https://peterkovesi.github.io/ImagePhaseCongruency.jl/dev/examples/#Fourier-transform-of-Moisan's-periodic-image-component-1>

Nate that I am now retired and find myself not doing very much coding these days.  Indeed at the moment I am often off the grid for extended periods of time. Thus I am not in a position to provide code reviews but should you have any questions about my code I will be more than happy to try to answer them.  In the meantime I will be cheering you on from the sidelines!

src/gaborkernels.jl
```diff
@@ -0,0 +1,323 @@
...
+    ω_denom, θ_denom = kern.ω_denom, kern.θ_denom
+    # Although in `getindex`, the computation is heavy enough that this runtime if-branch is
+    # harmless to the overall performance at all
+    if kern.normalize
+        # normalize: from reference [1] of LogGabor
+        # By changing division to multiplication gives about 5-10% performance boost
+        x, y = (x, y) .* kern.freq_scale
+    end
+    ω = sqrt(x^2 + y^2) # this is faster than hypot(x, y)
+    θ = atan(y, x)
```
johnnychen94, line 226: It is often the case that we can pre-compute the frequency `ω` and orientation `θ` and then call `kern[ω, θ]`. In this case, we might want to introduce a sphere index type to cache the computation and make the `getindex` dispatch on it.

```julia
for p in SphereIndices(img)
    # by doing this, we only need to compute the sphere index once
    kern1[p] * kern2[p]
end
```

Because some kernels are defined orientation-invariant, they wouldn't need the `θ` information at all. Thus each component of `SphereIndex` must be computed lazily so that we don't compute `θ`.

cc @Tokazama for a possible use case for https://github.com/JuliaArrays/SpatioTemporalTraits.jl/pull/11. This is kind of the opposite of the `CompositeSystem` you proposed.

johnnychen94:
@peterkovesi Thanks for the information! I've made an initial implementation of the `phasecong3` and it turns out that I only get 2x slower 😂 because of the repeated computation as I noted in https://github.com/JuliaImages/ImageFiltering.jl/pull/235#discussion_r741936206. The idea of separating the log Gabor frequency filter and the log Gaussian angular filter [here](https://github.com/peterkovesi/ImagePhaseCongruency.jl/blob/9cc9dbddccb741c9bed7b3b6c44436b797cf32c6/src/phasecongruency.jl?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L1348) to decouple the orientation and scale loops is really really insightful.

Profiling my initial implementation code shows that almost 80% of the computation is spent on the log Gabor filter, which is definitely a performance hotspot.

That said, this PR is far from ready or satisfying.

src/gaborkernels.jl
Tokazama: I've long wanted to be able to have a method for explicitly extracting spatial indices and this is a great practical example that motivates an implementation. We need some way of having a spatial system in the midst of color, time, etc. which users may end up creating a view of. The `CompositeSystem` proposal was an attempt to accommodate that, but if we did go in that direction we would still need to be able to extract a succinct representation of just the spatial system (e.g., `SphericalSystem`).

timholy: We do already have `coords_spatial` and `ImageAxes.filter_space_axes`. I'm not a fan of color axes because having one destroys some of our most important abstractions (one array element = one pixel), but I recognize we may want to support such constructs slightly better than we do now.

src/gaborkernels.jl
```diff
@@ -0,0 +1,323 @@
...
+        expected_size = @. 5λ * length(ax)
+        @warn "for wavelength `λ=$λ` the expected kernel size is expected to be larger than $expected_size."
+    end
+
+    Gabor{Complex{T}, T, typeof(first(ax))}(ax, params...)
+end
+
+@inline Base.size(kern::Gabor) = map(length, kern.ax)
+@inline Base.axes(kern::Gabor) = kern.ax
+@inline function Base.getindex(kern::Gabor, x::Int, y::Int)
```
timholy, line 161: Don't we index in `y, x` format?

src/gaborkernels.jl
```diff
@@ -0,0 +1,323 @@
...
+    ψ, σ, γ = kern.ψ, kern.σ, kern.γ
+    s, c = kern.sc # sincos(θ)
+    λ_scaled = kern.λ_scaled # 2π/λ
+
+    xr = x*c + y*s
+    yr = -x*s + y*c
+    yr = γ * yr
+    return exp(-(xr^2 + yr^2)/(2σ^2)) * cis(xr*λ_scaled + ψ)
```
timholy, line 169: What happens if we index this with out-of-bounds `x` and `y`? If lack of `BoundsError` is deliberate, it should be documented.

timholy (commented):
This is very nice! One overall comment (multidimensional police reporting for duty, sir :smile:): how about parametrizing this as in https://stackoverflow.com/a/25633868/1939814? You can of course rotate the coordinates before applying that formula, so it might make sense to also allow a rotation transformation.

src/gaborkernels.jl
Tokazama: > We do already have `coords_spatial` and `ImageAxes.filter_space_axes`.

I thought those methods were only reliable if the type encoding the coordinate system is kept at the top level and it was a cartesian system, like we usually assume with AxisArray. Could we rely on that approach if the coordinate system was encoded in nested arrays or the spatial indices couldn't just be permuted or sliced (e.g., one dimension was for the radius and the other was for the azimuth)?

>  I'm not a fan of color axes because having one destroys some of our most important abstractions 

I agree, but it would be good to play nice with the machine learning community on that one.

timholy: We can fix the nesting in the same way you've been doing it in ArrayInterface. Don't need a new trait that does the same thing, right?

timholy: But yeah, we should provide better support for ML. Not by changing our algorithms, but by making it easier to represent their raw data in our framework. Making algorithms work for arrays that have a color channel is messy and error-prone, because there is no way to know how to interpret that channel unless it's annotated unambiguously. In which case, why not just change the representation to get rid of the color channel?

Tokazama: So if we wanted to facilitate what was proposed we could do something like this

```julia
SpatialIndices(x) = SpatialIndices(CoordinateSystem(x), x)
SpatialIndices(cs::SphericalSystem, x) = SphereIndices{SphereIndex}(cs, indices_spatial(x))
```

We still use existing traits (`indices_spatial`), but we need more information to know which indices correspond to the radius, azimuth, and zenith so we can compose each `SphereIndex`.

johnnychen94: Because some of the computation can be non-trivial, can this design handle the task "I only want to iterate the first dimension of the system so if possible, don't do the extra computation for the second dimension"?

Tokazama: If I'm understanding everything correctly here, you have two unique sets of coordinate systems with corresponding spatial indices/coordinates.
I would think that's sufficient to have a function like this:
```julia
function map_spatial_indices(dest, dest_cs::CartesianSystem, src, src_cs::SphericalSystem)
    ...
end
```
That function certianly isn't pretty and it would be nice if we could make it work through `getindex`.

MendeBadra:
What happened to this rework of Gabor function in ImageFiltering.jl? When I google this, it returns this stuff but not implemented in ImageFiltering.jl?

PR #236, opened 2021-11-02
Reexport FFTW

johnnychen94:
`fft`/`ifft` and related utils might be something people would want to have when they `using ImageFiltering`.

However, I don't have a strong opinion here; actually, I'm trying to be conservative on exporting too many symbols.

Thoughts?

### DIFF
--- a/src/ImageFiltering.jl
+++ b/src/ImageFiltering.jl
@@ -1,6 +1,6 @@
 module ImageFiltering
 
-using FFTW
+@reexport using FFTW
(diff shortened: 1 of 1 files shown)


timholy:
I'm unsure, but on balance maybe not necessary? Is there a reason that someone using this would have to have the same `fft` that ImageFiltering uses internally? What if a pure-Julia implementation comes along?

johnnychen94:
Okay, let's close this for now.

PR #238, opened 2021-11-06
add ROF ADMM solver

johnnychen94:
This is currently incomplete, the benchmark shows that it's already faster than the MATLAB equivalent version. Yet I think we still have more room for optimization wrt the FFT-related operations, so I opened this draft to see if there are any suggestions on this.

```julia
using ImageFiltering, ImageCore, TestImages, ImageShow
using ImageFiltering.Models

img = Float32.(testimage("cameraman"))
noisy_img = img .+ 0.25 .* randn(size(img))

@btime solve_ROF_ADMM($noisy_img, 0.25, 0.02; num_iters=100);
# julia + FFTW v1.1.1: 856.001 ms (104403 allocations: 639.15 MiB)
# julia + FFTW v1.4.5: 1.275 s (8316 allocations: 632.49 MiB)
# matlab: 1.851779s
```

FFTW has some performance regression here: https://github.com/JuliaMath/FFTW.jl/issues/224

### DIFF
--- a/src/models.jl
+++ b/src/models.jl
@@ -13,10 +13,12 @@ This submodule provides predefined image-related models and its solvers that can
 by many image processing tasks.
 
 - solve the Rudin Osher Fatemi (ROF) model using the primal-dual method: [`solve_ROF_PD`](@ref) and [`solve_ROF_PD!`](@ref)
+- solve the Rudin Osher Fatemi (ROF) model using the ADMM method: [`solve_ROF_ADMM`](@ref) and [`solve_ROF_ADMM!`](@ref)
 """
 Models
 
-export solve_ROF_PD, solve_ROF_PD!
+export solve_ROF_PD, solve_ROF_PD!,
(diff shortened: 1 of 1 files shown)


src/models.jl
```diff
@@ -169,5 +171,129 @@ function _l2norm_vec!(out, Vs::Tuple)
...
+        @. RHS = λ/μ * f
+        for i in 1:length(d)
+            # negative adjoint gradient is the reverse finite difference
+            @. Δbd = b[i] - d[i]
+            # reuse d here as buffer as it will be updated later in d-subproblem
+            RHS .+= fdiff!(d[i], Δbd; dims=i, rev=true)
+        end
+        # TODO(johnnychen94): optimize the memory allocation
+        fft_tmp .= μ .* fft(RHS)./LHS
+        u .= real.(ifft(fft_tmp))
```
johnnychen94, line 271: This is the performance hot spot:



I guess we can use in-place version `fft!` or real version `rfft` to boost it.

Issue #240, opened 2021-11-24
Butterworth Kernel?


babaq:
Hi,

I can't find a butterworth kernel in the Images, and think would be a nice addition to the package.

The `DSP.jl` do have a Butterworth filter design, maybe could be used or looked up to generate the Factored Kernels. 

Alex

PR #242, opened 2022-01-17
Add Butterworth Filter

ashwinn-v:
In response to feature request #240 
Please let me know if there are any fixes to be done on the code.


src/kernel.jl
```diff
@@ -509,6 +509,15 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
     sum(x.^2)
 end
 
+function butterworth(n::Real,Wn::Real,ls::Tuple{Integer, Integer})
```
johnnychen94, line 512: The style should be consistent, you can additionally check out https://github.com/invenia/BlueStyle (but try to keep the same to the current codebase you're working on)

```suggestion
function butterworth(n::Real, Wn::Real, ls::Tuple{Integer,Integer})
```

johnnychen94 (changes requested):
Docs and tests are needed.

Some docstrings in ImageFiltering are written in Julia pre-1.0 age, so you can follow the guideline in https://docs.julialang.org/en/v1/manual/documentation/#man-documentation. I personally prefer a verbose docstring, e.g., the docstring for `Gabor` in #235. You can also follow how unit tests are written in that PR.

src/kernel.jl
```diff
@@ -509,6 +509,26 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
...
+butterworth(n,Wn,ls{tuple}) -> k
+Returns a multidimensional dimensional Butterworth kernel contained in an OffsetArray(::Matrix{Float64})
+
+    - `n` is the order of the filter
+    - `Wn` is the normalized cutoff frequency
+    - `ls` is the size of the kernel
+#Citation
+Selesnick, Ivan W., and C. Sidney Burrus. "Generalized digital Butterworth filter design." IEEE Transactions on signal processing 46.6 (1998): 1688-1694.
+"""
+
```
johnnychen94, line 522: Extra empty line here would cause `?Kernel.butterworth` giving nothing.

```suggestion
```

src/kernel.jl
```diff
@@ -509,6 +509,26 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
     sum(x.^2)
 end
 
+"""
+butterworth(n,Wn,ls{tuple}) -> k
```
johnnychen94, line 513: indentation, and `::` for type annotation.

```suggestion
    butterworth(n, Wn, ls::Dims) -> k
```

src/kernel.jl
```diff
@@ -509,6 +509,26 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
...
+Returns a multidimensional dimensional Butterworth kernel contained in an OffsetArray(::Matrix{Float64})
```
johnnychen94, line 514: one extra new line after the signature.

```suggestion

Returns a multidimensional dimensional Butterworth kernel contained in an OffsetArray(::Matrix{Float64})
```

src/kernel.jl
```diff
@@ -509,6 +509,26 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
...
+
+    - `n` is the order of the filter
+    - `Wn` is the normalized cutoff frequency
+    - `ls` is the size of the kernel
```
johnnychen94, line 518: You don't need indentation for the markdown list

```suggestion
- `n` is the order of the filter
- `Wn` is the normalized cutoff frequency
- `ls` is the size of the kernel
```

johnnychen94, line 518: why you use `ls` for kernel size instead of `sz`? The former seems to come from the math equation of butterworth kernel. Can you also add the equation to the docstring so that people immediately knows what `n`, `Wn`, `ls` means and how they affects the results?

src/kernel.jl
```diff
@@ -509,6 +509,26 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
...
+"""
+
+function butterworth(n::Real, Wn::Real, ls::Tuple{Integer,Integer})
+    ws = map(n->(ceil(Int,n)>>1), ls)
+    R = CartesianIndices(map(w->IdentityUnitRange(-w:w), ws))
+    nn = 2*n
+    @. 1/(1+(sqrt(2)-1)*(df(R)/Wn)^nn)
+end   
+
+butterworth(n::Real, Wn::Real, ls::Integer)    = butterworth(n, Wn, (ls,ls))
```
johnnychen94, line 530: This convenient method only makes sense when `butterworth` is 2D only; otherwise, we should not add it.

test/specialty.jl
```diff
@@ -238,6 +238,14 @@ using ImageFiltering: IdentityUnitRange
         fwhm = ceil(Int, (α*2*sqrt(2^(1/β) - 1)) * 4)
         @test Kernel.moffat(α, β) == Kernel.moffat(α, β, (fwhm, fwhm))
     end
+
+    @testset "butterworth" begin
+        a = rand()
+        b = rand()
+        @test Kernel.butterworth(a,b,(3,3)) == Kernel.butterworth(a,b,3)
+        @test Kernel.butterworth(a,b,(4,4)) == Kernel.butterworth(a,b,4)
```
johnnychen94, line 246: We also need some numerical tests here, can you test against one or two results generated from other frameworks that has `butterworth` implemented or manual calculated results? Also do remember to add a comment on how the references are generated.

We also need to test that the kernel axis are "centered". For instance, the axes should be `(-1:1, -1:1)` when kernel size is `(3, 3)`. Also need to test the even case as well.

src/kernel.jl
```diff
@@ -509,6 +509,26 @@ moffat(α::Real, β::Real)                 = moffat(α, β, ceil(Int, (α*2*sqrt
...
+Returns a multidimensional dimensional Butterworth kernel contained in an OffsetArray(::Matrix{Float64})
+
+    - `n` is the order of the filter
+    - `Wn` is the normalized cutoff frequency
+    - `ls` is the size of the kernel
+#Citation
+Selesnick, Ivan W., and C. Sidney Burrus. "Generalized digital Butterworth filter design." IEEE Transactions on signal processing 46.6 (1998): 1688-1694.
+"""
+
+function butterworth(n::Real, Wn::Real, ls::Tuple{Integer,Integer})
```
johnnychen94, line 523: It looks like if we do

```suggestion
function butterworth(n::Real, Wn::Real, ls::NTuple{N,Integer}) where N
```

we immediately get a N-dimensional implementation as the docstring says: "multidimensional dimensional Butterworth kernel". Can you check this and if it works, add 1d and 3d tests? Also don't forget to update the docstring `OffsetArray(::Matrix{Float64})`.

PR #244, opened 2022-02-08
Less alloc in TriggsSdika filtering (fix #243)

### LINKED ISSUE #243: `imfilter!` allocates significantly
I believe this setup, with `imfilter!` and `symmetric` should be allocation-free/minimal?

Below uses the new allocation profiler in julia 1.8

```
julia> img1 = rand(Float32, 1000, 1000);

julia> img2 = similar(img1);

julia> kern = KernelFactors.IIRGaussian((2.0,2.0));

julia> @time imfilter!(img2, img1, kern, "symmetric");
  1.017254 seconds (2.70 M allocations: 144.354 MiB, 3.17% gc time, 97.60% compilation time)

julia> @time imfilter!(img2, img1, kern, "symmetric");
  0.017806 seconds (1.03 k allocations: 7.768 MiB)

julia> using Profile, PProf

julia> Profile.Allocs.@profile sample_rate=0.5 imfilter!(img2, img1, KernelFactors.IIRGaussian((2.0,2.0)), "symmetric");

julia> PProf.Allocs.pprof(from_c = false)
```
Screenshot from 2022-02-06 22-49-56

cc. @johnnychen94

timholy:
Putting the string interpolation behind a `@noinline` is the key change,
but it doesn't hurt to hoist the range creation.

Of the few remaining allocations, about half will be fixed by
https://github.com/JuliaLang/julia/pull/43990, as they are another
instance of https://github.com/JuliaLang/julia/issues/35972

### DIFF
--- a/src/imfilter.jl
+++ b/src/imfilter.jl
@@ -1327,7 +1327,7 @@ _imfilter_inplace_tuple!(r, out, img, ::Tuple{}, Rbegin, inds, Rend, border) = o
     if length(ind) <= max(k, l)
         throw_imfilter_dim(Rbegin, length(ind), max(k, l))
     end
-    indleft = ind[begin:begin+k-1]
+    indleft = ind[firstindex(ind):firstindex(ind)+k-1]
     indright = ind[end-l+1:end]
     for Iend in Rend
         # Initialize the left border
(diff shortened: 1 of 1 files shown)


timholy:
CC @IanButterworth

IanButterworth:
Many thanks! Though I thought this would have a bigger effect
```
julia> @btime imfilter!(img2, img1, kern, "symmetric");
  19.865 ms (30 allocations: 7.69 MiB)
```
master
```
julia> @btime imfilter!(img2, img1, kern, "symmetric");
  20.384 ms (1031 allocations: 7.77 MiB)
```

Also, I don't quite understand why the key change here makes a difference.. if that `if` branch isn't hit, were the interpolations still done when inlined??!

timholy:
It somehow triggers retention and propagation of a `gcframe`. I don't really understand the details, unfortunately.

PR #247, opened 2022-04-26
correct docstrings on kernel.gaussian syntax

Iddingsite:
Remove brackets in docstring for kernel length, as it requires a tuple and not a array.

### DIFF
--- a/src/kernel.jl
+++ b/src/kernel.jl
@@ -259,7 +259,7 @@ bickley() = product2d(KernelFactors.bickley())
 bickley(extended, d) = (broadcast(*, KernelFactors.bickley(extended, d)...),)
 
 """
-    gaussian((σ1, σ2, ...), [(l1, l2, ...)]) -> g
+    gaussian((σ1, σ2, ...), (l1, l2, ...)) -> g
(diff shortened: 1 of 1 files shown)


src/kernel.jl
```diff
@@ -259,7 +259,7 @@ bickley() = product2d(KernelFactors.bickley())
 bickley(extended, d) = (broadcast(*, KernelFactors.bickley(extended, d)...),)
 
 """
-    gaussian((σ1, σ2, ...), [(l1, l2, ...)]) -> g
+    gaussian((σ1, σ2, ...), (l1, l2, ...)) -> g
```
johnnychen94, line 262: This is an undocumented manpage convention that anything listed in `[]` is optional, thus here it indicates two usage:

```julia
gaussian((σ1, σ2, ...), (l1, l2, ...)) -> g
gaussian((σ1, σ2, ...)) -> g
```

see also https://stackoverflow.com/questions/23242493/linux-unix-man-page-syntax-conventions

We should probably simplify it to `gaussian(σ, [l])` and document that both `σ` and `l` can be a tuple.

PR #255, opened 2022-12-28
Use SnoopPrecompile for precompilation

timholy:
This updates us to the latest options in precompilation. On my machine, the initial invocation of

```
img = rand(Float32, 100, 100)
@time @eval imfilter(img, KernelFactors.gaussian((1.5f0, 1.5f0)));
```

drops (on Julia nightly) from 0.78s to 0.007s, a 100x improvement.

closes https://github.com/JuliaImages/ImageQualityIndexes.jl/issues/53


src/precompile.jl
```diff
@@ -1,36 +1,51 @@
...
+        rand(Float32, 100, 100),
+        rand(Float64, 100, 100),
+        rand(Gray{N0f8}, 100, 100),
+        rand(Gray{N0f16}, 100, 100),
+        rand(RGB{N0f8}, 100, 100),
+    ]
+    images3d = Any[
+        rand(Gray{N0f16}, 100, 100, 10),
+    ]
+    # We could put the kernels and resources here too, but it makes sense to precompile them
```
johnnychen94, line 14: I'm not sure if I understand the comment. Should we precompile other kernels, or it's good enough to only take Gaussian as a hint to generate precompilation codes? I assume you mean the latter?

timholy: Only the things that are inside `@precompile_all_calls` are guaranteed to be precompiled along with "all" the things they depend on. (There are a few caveats, but that's pretty close to accurate.) So this only caches specializations for the supplied object types. If there are more you think we should precompile, we can add them, at the cost of possibly increasing compilation time.

It's nearly free, though, to precompile additional "user-level" code that funnels to a common pipeline. For example, if we have 20 different kernel generators, but they all return a `SMatrix{5, 5, Float64}`, then the `imfilter` code will only be compiled once for `SMatrix{5, 5, Float64}` kernels even if we generate those kernels with many different high-level functions. So in that case, increasing the number of kernels we compile barely increases compilation time at all.

Hence, it's not unreasonable to add more things to this list if you want them.

PR #263, opened 2023-07-20
[compat] ImageCore 0.10

timholy:


timholy:
Failure on 1.0 is due to test-circular dependencies with ImageQualityIndexes.

PR #264, opened 2023-07-21
Drop Julia 1.0, ImageMagick

timholy:
The previous version seems to run on Julia 1.0, but there's a conflict
installing test-only dependencies. Therefore it's best to drop 1.0.

This also eliminates ImageMagick as a test dependency.

Finally, it fixes the `ColorVectorSpace.Future.abs2` depwarn, which was killing performance.


timholy:
Doc failure will be fixed by https://github.com/JuliaDocs/DemoCards.jl/pull/147

Issue #265, opened 2023-08-07
Incorrect usage of `threadid`


cafaxo:
This usage of `threadid` definitely does not look correct:
https://github.com/JuliaImages/ImageFiltering.jl/blob/afd33f4e2903bc29f37cfe38afc2b7d7ff1142d8/src/imfilter.jl#L487-L493
See https://julialang.org/blog/2023/07/PSA-dont-use-threadid/#the_buggy_pattern

mkitti:
This probably should at least use the static scheduler if not be refactored.

IanButterworth:
I would refactor it. Using the static scheduler isn't a great stopgap re. composability

PR #266, opened 2023-08-09
replace bad threadid buffering

IanButterworth:
Fixes https://github.com/JuliaImages/ImageFiltering.jl/issues/265

This is a naive swap out. There may be refactoring that should happen at a higher level with the buffer management.

Untested locally.


timholy (approved):
Thanks, looks fine to me!

PR #271, opened 2024-01-13
Enable reusing fft plans

IanButterworth:
Requires 
- https://github.com/HolyLab/RFFT.jl/pull/17

TODO:
- [x] Merge https://github.com/HolyLab/RFFT.jl/pull/17
- [x] Register RFFT.jl in General
- [x] Tidy (remove manual RFFT add, remove demo)
- [x] Settle on naming/API i.e. should this be setup via `Algorithm.FFT(img, kernel, border)`?
- [x] Docstrings

___

Because they are not threadsafe FFTW FFT plans are behind a lock. They should be reusable for same-sized images and kernels, so this is WIP towards exposing a way to provide plans, so that `imfilter!` can be run concurrently on julia threads without lock conflict. 

Note that this is using https://github.com/JuliaLang/julia/pull/52883 to count conflicts.

There's a demo temporarily in this PR.

Before it hits 34255 lock conflicts
```
% FFTW_NUM_THREADS=1 BLAS_NUM_THREADS=3 ../julia/julia --project -t6,1 -q
julia> include("demo.jl")
warm run of benchmark(mats): 5.483911 seconds (2.14 M allocations: 10.671 GiB, 16.62% gc time, 34255 lock conflicts)
```

This PR removes all but the lock conflicts from the singlular plans, which could be moved out of the for loop
```
% FFTW_NUM_THREADS=1 BLAS_NUM_THREADS=3 ../julia/julia --project -t6,1 -q
julia> include("demo.jl")
warm run of benchmark(mats): 2.440659 seconds (478.12 k allocations: 7.888 GiB, 21.53% gc time, 8 lock conflicts)
```


IanButterworth:
Seeing as this seems to work, what would be a good design to pass this through the API?

Given that plans are an FFT peculiarity I wasn't sure if there was an appropriate way to pass it through the more generic `imfilter!`methods.

IanButterworth:
Maybe this would make sense?
```
r = CPU1(Algorithm.FFT(img, kern))
imfilter!(r, buffer, img, kern)
```

So make a constructor that generates plans if args are given?
https://github.com/JuliaImages/ImageFiltering.jl/blob/08e34f2b7025c51b0ba9aab9249fac11d3897fe5/src/ImageFiltering.jl#L56

src/imfilter.jl
```diff
@@ -840,12 +840,33 @@ function _imfilter_fft!(r::AbstractCPU{FFT},
...
+    bord = border(kernel, A, Algorithm.FFT())
+    _A = padarray(T, A, bord)
+    p1 = plan_rfft(_A)
+    kern = samedims(_A, kernelconv(kernel...))
+    krn = FFTView(zeros(eltype(kern), map(length, axes(_A))))
+    p2 = plan_rfft(krn)
+    B = p1 * _A
+    B .*= conj!(p2 * krn)
+    p3 = plan_irfft(B, length(axes(_A, 1)))
+    return Algorithm.FFT(p1, p2, p3)
```
IanButterworth, line 855: Can any of these actual operations be replaced with just constructing things based on dims etc.?

IanButterworth:
Ok, I think this seems reasonable now. @timholy would you mind reviewing for general design and I can add tests & docs if you agree

timholy:
I confess I haven't thought about these plans in ages (the code I linked earlier was probably written in the Julia 0.1 days), but nothing bad jumps out about this. Thanks for doing it!!

src/ImageFiltering.jl
```diff
@@ -50,10 +50,16 @@ function Base.transpose(A::StaticOffsetArray{T,2}) where T
 end
 
 module Algorithm
+    import FFTW
     # deliberately don't export these, but it's expected that they
     # will be used as Algorithm.FFT(), etc.
     abstract type Alg end
-    "Filter using the Fast Fourier Transform" struct FFT <: Alg end
+    "Filter using the Fast Fourier Transform" struct FFT <: Alg
+        plan1::Union{FFTW.rFFTWPlan,Nothing}
```
JeffFessler, line 58: Probably `rFFTWPlan` is not a concrete type?  So should this be a parametric struct?
(I might be using the terms incorrectly...)

IanButterworth: This PR isn't in a working state yet. My plan is to bring the in-place functions over from RFFT.jl

timholy: If one can move RFFT from one registry to another, I'm happy to do that if it would help.

IanButterworth: Thanks but I was going to suggest that once/if I got it working through just copying the code into here first

IanButterworth:
I'm a little stuck.

Currently running `demo.jl` gives
```
julia> include("demo.jl")
ERROR: LoadError: BoundsError: attempt to access 49×96 Matrix{ComplexF32} at index [CartesianIndex(81, 87)]
Stacktrace:
  [1] throw_boundserror(A::Matrix{ComplexF32}, I::Tuple{CartesianIndex{2}})
    @ Base ./abstractarray.jl:734
  [2] checkbounds
    @ ./abstractarray.jl:699 [inlined]
  [3] macro expansion
    @ ./multidimensional.jl:1128 [inlined]
  [4] copyto!(dest::Matrix{Float32}, Rdest::CartesianIndices{2, Tuple{…}}, src::Matrix{ComplexF32}, Rsrc::CartesianIndices{2, Tuple{…}})
    @ Base ./multidimensional.jl:1119
  [5] _imfilter_fft!(r::CPU1{…}, out::Matrix{…}, A::OffsetArrays.OffsetMatrix{…}, kernel::Tuple{…}, border::NoPad{…})
    @ ImageFiltering ~/Documents/GitHub/ImageFiltering.jl/src/imfilter.jl:0
  [6] imfilter!
    @ ImageFiltering ~/Documents/GitHub/ImageFiltering.jl/src/imfilter.jl:816 [inlined]
  [7] imfilter!
    @ ImageFiltering ~/Documents/GitHub/ImageFiltering.jl/src/imfilter.jl:340 [inlined]
  [8] imfilter!(r::CPU1{…}, out::Matrix{…}, img::SubArray{…}, kernel::Tuple{…}, border::Pad{…})
    @ ImageFiltering ~/Documents/GitHub/ImageFiltering.jl/src/imfilter.jl:327
  [9] imfilter!(r::AbstractResource, out::AbstractArray{…}, img::AbstractArray{…}, kernel::Tuple, border::Union{…}) where {…}
    @ ImageFiltering ~/Documents/GitHub/ImageFiltering.jl/src/imfilter.jl:217 [inlined]
 [10] test(mats::Vector{Array{Float32, 3}})
    @ Main ~/Documents/GitHub/ImageFiltering.jl/demo.jl:31
```

For the existing approach, `irfft` call takes `size(B) = (49, 96)` and returns a `(96, 96)` matrix

https://github.com/JuliaImages/ImageFiltering.jl/blob/a6a3cce826b891d979cee3668fb7177f52683e45/src/imfilter.jl#L884


For the new approach with `buffered_planned_irfft` `RFFT.RCpair{T}(undef, size(a))` takes `size(a) = (96, 96)` and creates a buffer with `(49, 96)` which means the output of its lambda is a `(49, 96)` array, where the remainder of `_imfilter_fft!` is expecting a `(96, 96)` matrix.
https://github.com/JuliaImages/ImageFiltering.jl/blob/a6a3cce826b891d979cee3668fb7177f52683e45/src/imfilter.jl#L851-L858

If this is clear to anyone, guidance would be appreciated. Otherwise I'll try to change `RFFT.RCpair` to accomodate, but I don't understand that constructor.

https://github.com/JuliaImages/ImageFiltering.jl/blob/a6a3cce826b891d979cee3668fb7177f52683e45/src/RFFT.jl#L15-L24

IanButterworth:
Comparing buffered vs standard I get different values here
https://github.com/JuliaImages/ImageFiltering.jl/blob/78044e7f3b85cde221fe41cd02f6a97e1af135fc/src/imfilter.jl#L875-L877
```
B[1:4] =  ComplexF32[3065.7727f0 + 113.279686f0im, -254.48924f0 - 85.513504f0im, -369.77383f0 - 146.0826f0im, -267.98865f0 - 201.43063f0im]
_B[1:4] = ComplexF32[4595.6045f0 + 0.0f0im,         3.6047888f0 + 15.441383f0im,  14.378552f0 - 19.552792f0im, -18.684448f0 - 19.284311f0im]
```

My trial and error phase of debugging isn't being very fruitful, so I think I'll wait for review. @timholy if you happen to have time

mkitti:
Commeting to remind myself to take a look

demo.jl
```diff
@@ -0,0 +1,61 @@
...
+        f1 = deepcopy(mat[:, :, 1])
+        r_cached = CPU1(ImageFiltering.planned_fft(f1, kernel))
+        f2 = deepcopy(mat[:, :, 1])
+        r_noncached = CPU1(Algorithm.FFT())
+        for i in axes(mat, 3)
+            frame = @view mat[:, :, i]
+            imfilter!(r_noncached, f2, frame, kernel)
+            imfilter!(r_cached, f1, frame, kernel)
+            @show f1[1:2, 1:2] f2[1:2, 1:2]
+            all(f1 .≈ f2) || error("f1 !≈ f2")
```
timholy, line 33:
```suggestion
            f1 ≈ f2 || error("f1 !≈ f2")
```
This form takes "scale" into account in assessing differences. Can be important in deciding whether something that should be zero (or very small) is close enough.

src/RFFT.jl
```diff
@@ -0,0 +1,75 @@
...
+using FFTW, LinearAlgebra
+
+export RCpair, plan_rfft!, plan_irfft!, rfft!, irfft!, normalization
+
+import Base: real, complex, copy, copy!
+
+mutable struct RCpair{T<:AbstractFloat,N,RType<:AbstractArray{T,N},CType<:AbstractArray{Complex{T},N}}
+    R::RType
+    C::CType
+    region::Vector{Int}
```
timholy, line 12: I guess this is called `dims` in modern parlance?

src/RFFT.jl
```diff
@@ -0,0 +1,75 @@
...
+    return RC
+end
+function irfft!(RC::RCpair{T}) where T
+    p = rplan_inv(RC.C, RC.R, RC.region, FFTW.ESTIMATE, FFTW.NO_TIMELIMIT)
+    FFTW.unsafe_execute!(p, RC.C, RC.R)
+    rmul!(RC.R, 1 / prod(size(RC.R)[RC.region]))
+    return RC
+end
+
+@deprecate RCpair(realtype::Type{T}, realsize, region=1:length(realsize)) where T<:AbstractFloat RCpair{T}(undef, realsize, region)
```
timholy, line 73: Is this new or a copy of a deprecation in RFFT.jl? If the latter, I think we could have a breaking release.

timholy (commented):
Nothing jumps out as a serious problem just from a read through the code, but it's probably been a decade since I put the serious work into writing RFFT.jl (it used to be part of a more monolithic repository so the git history may not reflect that). Do the disagreements only arise when you use multiple threads?

src/RFFT.jl
IanButterworth: It's a copy

IanButterworth:
I'm stuck.

So this sanity check evaluates true
```
using RFFT, FFTW
a = rand(Float64, 10, 10)
out1 = rfft(a)

buf = RFFT.RCpair{Float64}(undef, size(a))
rfft_plan = RFFT.plan_rfft!(buf)
copy!(buf, a)
out2 = complex(rfft_plan(buf))

out1 ≈ out2
```

however, for equal `krn` inputs (validated through typeof and sum equality) these lines give different results

## The existing uncached plan approach

https://github.com/JuliaImages/ImageFiltering.jl/blob/a79e8078d3adc4481db3b66ce5990b390bddecc7/src/imfilter.jl#L915
```
┌ Info: good one
│   (rfft(krn))[1:3] =
│    3-element Vector{ComplexF64}:
│     -0.00014399948754237335 + 0.0im
│       -0.004414546165525474 + 1.4913814981424873e-17im
└       -0.017117085642739213 + 8.853611881175574e-18im
```

## Cached plan approach
https://github.com/JuliaImages/ImageFiltering.jl/blob/a79e8078d3adc4481db3b66ce5990b390bddecc7/src/imfilter.jl#L893
```
┌ Info: bad one 1
│   (complex(planned_rfft2(AbstractFFTs.realfloat(krn))))[1:3] =
│    3-element Vector{ComplexF64}:
│     -0.00014399948754237335 + 0.0im
│       -0.004405094283318256 - 0.0002887251333562888im
└       -0.016970646602189325 - 0.0022342280108512223im
```

where `planned_rfft1` is generated from
https://github.com/JuliaImages/ImageFiltering.jl/blob/a79e8078d3adc4481db3b66ce5990b390bddecc7/src/imfilter.jl#L845-L852

Project.toml
```diff
@@ -4,6 +4,7 @@ author = ["Tim Holy <tim.holy@gmail.com>", "Jan Weidner <jw3126@gmail.com>"]
 version = "0.7.8"
 
 [deps]
+AbstractFFTs = "621f4979-c628-5d54-868e-fcf4e3e8185c"
```
timholy, line 7: Also need a [compat] entry for this.

demo.jl
```diff
@@ -0,0 +1,67 @@
+using ImageFiltering, FFTW, LinearAlgebra, Profile, Random
```
timholy, line 1: I saw the "temporarily" comment in the conversation, so I'm assuming this will change.

src/imfilter.jl
```diff
@@ -837,13 +837,61 @@ function _imfilter_fft!(r::AbstractCPU{FFT},
...
+    return out
+end
+
+function buffered_planned_rfft(a::AbstractArray{T}) where {T}
+    buf = RFFT.RCpair{T}(undef, size(a))
+    plan = RFFT.plan_rfft!(buf; flags=FFTW.MEASURE)
+    return function (arr::AbstractArray{T}) where {T}
+        copy!(buf, OffsetArrays.no_offset_view(arr))
+        return plan(buf)
+    end
```
timholy, line 849: No test coverage here and below

test/gabor.jl
```diff
@@ -7,11 +7,15 @@ using ImageFiltering, Test, Statistics
...
-    @test isequal(size(kernel[1]),(size_x,size_y))
+
+    @test_logs (:warn, r"The input parameter size_") match_mode=:any begin
+        kernel = Kernel.gabor(0,0,σx,0,5,γ,0)
+        @test isequal(size(kernel[1]),(size_x,size_y))
+    end
+    @test_logs (:warn, r"The input parameter size_") match_mode=:any begin
+        kernel = Kernel.gabor(0,0,σx,π,5,γ,0)
+        @test isequal(size(kernel[1]),(size_x,size_y))
+    end
```
timholy, line 18: Thanks!!

test/nd.jl
```diff
@@ -50,6 +50,8 @@ Base.zero(::Type{WrappedFloat}) = WrappedFloat(0.0)
     v = fill(0xff, 10)
     kern = centered(fill(0xff, 3))
     @info "Two warnings are expected"
+    # TODO: use @test_logs (:warn, r"Likely overflow or conversion error detected") match_mode=:any
+    # julia has an internal error currently on 1.10.2 that this hits https://github.com/JuliaLang/julia/pull/50759
```
timholy, line 54: 👍

src/imfilter.jl
```diff
@@ -837,13 +837,61 @@ function _imfilter_fft!(r::AbstractCPU{FFT},
...
+function filtfft(A, krn, planned_rfft1::Function, planned_rfft2::Function, planned_irfft::Function)
+    B = complex(planned_rfft1(A))
+    B .*= conj!(complex(planned_rfft2(krn)))
+    return real(planned_irfft(complex(B)))
+end
+# TODO: this does not work. See TODO below
+function filtfft(A::AbstractArray{C}, krn, planned_rfft1::Function, planned_rfft2::Function, planned_irfft::Function) where {C<:Colorant}
+    Av, dims = channelview_dims(A)
+    kernrs = kreshape(C, krn)
+    B = complex(planned_rfft1(Av, dims)) # TODO: dims is not supported by planned_rfft1
```
timholy, line 885: I don't think `dims` can be a point of flexibility: these plans are specific to the memory layout of the array. (The planning explores various implementations and picks the fastest discovered; performance is strongly dependent on memory layout, so the choice for one layout may not be the same as another.) You'd have to create a plan specifically to the colorant array-type.

timholy (commented):
Thanks for this! Not only is the new functionality appreciated, but the various cleanups are too.

I haven't played with this myself, but does `demo` pass now? If so, it seems like you're almost over the hump. I don't see anything objectionable here (assuming you'll get to a point of full test coverage for the new functionality), so feel free to run with this.

src/imfilter.jl
IanButterworth: I've added this comment to the TODO. I think I'd prefer to fix this in a follow on PR, if that sounds ok

PR #274, opened 2024-03-11
Fix docs build

IanButterworth:
@johnnychen94 there seems to be an issue with DemoCards.jl creating an index.md that cannot be found


ViralBShah:
Fixed in #284 by disabling the edit URLs generated by DemoCards which do not use the `src` directory. cc @mortenpi who may know how to fix this in DemoCards.jl.

ViralBShah:
I wish I had seen this PR earlier - since it would have saved me some other hunting effort...

PR #279, opened 2025-02-28
Fixes #278: LoG scale normalization

### LINKED ISSUE #278: Wrong LoG scale normalization in `blob_LoG`
The correct scale normalization for LoG should be `-σ^2` instead of `-σ` in the `multiLoG` function called by `blob_LoG`.

In addition, the docstring of `BlobLoG` says:
>radius is equal to σ√2

but
>radius is equal to σ√N for a N-dimensional image

is more accurate.

This error is easy to fix and I can handle it on my own (please note that this will be my first PR so any guidance is much appreciated!).

gschivre:
Simply replace `-σ` by `-σ^2` in `multiLoG` and update the docstring of `BlobLoG`.

### DIFF
--- a/src/extrema.jl
+++ b/src/extrema.jl
@@ -6,7 +6,7 @@ It has fields:
 - σ: the value of σ which lead to the largest `-LoG`-filtered amplitude at this location
 - amplitude: the value of the `-LoG(σ)`-filtered image at the peak
 
-Note that the radius is equal to σ√2.
+Note that the radius is equal to σ√N for a N-dimensional image.
 
 See also: [`blob_LoG`](@ref).
 """
@@ -96,7 +96,7 @@ function multiLoG(img::AbstractArray{T,N}, sigmas, σshape) where {T,N}
     @inbounds for (isigma, σ) in enumerate(sigmas)
         LoG_slice = @view img_LoG[isigma, colons...]
         imfilter!(LoG_slice, img, Kernel.LoG(ntuple(i->σ*σshape[i], Val(N))), "reflect")
-        LoG_slice .*= -σ
+        LoG_slice .*= -σ^2
     end
     return img_LoG
 end


ViralBShah:
Would be nice for someone else in the project to chime in to get this merged.

timholy:
I didn't work through this in detail, but are you sure this is correct? Gaussian densities have dimensions of the inverse of volume, and this seems like you might have changed this to inverse volume-squared?

gschivre:
> I didn't work through this in detail, but are you sure this is correct? Gaussian densities have dimensions of the inverse of volume, and this seems like you might have changed this to inverse volume-squared?

From my understanding, the amplitude response of the n-th derivative of a Gaussian scales with `σ^n`, so for the Laplacian of Gaussian (sum of second order derivative) it should be `σ^2`.

This can be checked:
```julia
using ImageFiltering, LinearAlgebra

# fake image with a single ball of radius r
function NDball(r, N = 2; size = ntuple(d -> 64, N), center = CartesianIndex(ntuple(d -> 32, N)))
    return [1.0 - (norm(c.I .- center.I) >= r) for c in CartesianIndices(size)]
end

# get the blob of highest amplitude
function getTopBlob(img, scales)
    blobs = blob_LoG(img, scales)
    return first(sort(blobs, by = b -> b.amplitude, rev = true))
end

getTopBlob(NDball(10, 1), 1:20)
#= With σ scalling
BlobLoG(location=CartesianIndex(32,), σ=(7,), amplitude=0.06165173598954085)
=#
#= With σ^2 scalling
BlobLoG(location=CartesianIndex(32,), σ=(10,), amplitude=0.4831344763518372)
=#

getTopBlob(NDball(10, 2), 1:20)
#= With σ scalling
BlobLoG(location=CartesianIndex(36, 24), σ=(1, 1), amplitude=0.3046499888904334)
=#
#= With σ^2 scalling
BlobLoG(location=CartesianIndex(32, 32), σ=(7, 7), amplitude=0.7355903572521065)
=#

getTopBlob(NDball(10, 3), 1:20)
#= With σ scalling
BlobLoG(location=CartesianIndex(40, 36, 32), σ=(1, 1, 1), amplitude=0.33406542371161835)
=#
#= With σ^2 scalling
BlobLoG(location=CartesianIndex(32, 32, 32), σ=(6, 6, 6), amplitude=0.9201386216134283)
=#

[getTopBlob(NDball(r, 1), 1:20).amplitude for r in 2:2:14]
#= With σ scalling
7-element Vector{Float64}:
 0.3989422804014327
 0.15830978051758404
 0.10687862635162675
 0.07780732143311218
 0.06165173598954085
 0.051051348521593064
 0.043324744637534995
=#
#= With σ^2 scalling
7-element Vector{Float64}:
 0.46352013527394104
 0.47888456067216945
 0.48169786073672155
 0.48268019904506626
 0.4831344763518372
 0.48331328709165605
 0.4816113967778672
=#
```
In all cases, the selected scale is wrong. Additionally, for a 1-D image, the amplitude decays with the radius with the `σ` scaling, but not with the `σ^2` scaling.

This is also the same scaling used by [scikit-image](https://github.com/scikit-image/scikit-image/blob/v0.25.2/skimage/feature/blob.py#L412-L581).

PR #285, opened 2025-05-07
feat(docs): color instead of boxes in padding

kunzaatko:
Color looks better and does not fill so much space. It makes the array
more compact looking unlike the boxes around every element.


kunzaatko:
Closing in favour of #287

PR #287, opened 2025-08-10
Remove unnecessary version check

kunzaatko:
- **feat(docs): color instead of boxes in padding**
- **fix: Remove check for Julia version**


kunzaatko:
Before:
image
After:
image

Before:
image
After:
image

Yeah sorry for that, actually this commit was not intended to be part of the PR. It is already in #285 and I got it mixed up. I am closing that in favour of this one.

kunzaatko:
In my opinion it is cleaner, but if you would suggest that it is not, I can revert that commit.

PR #289, opened 2025-08-11
Add const declerations to type aliases

kunzaatko:
- **feat(perf): Add const declarations to type aliases**

### DIFF
--- a/src/ImageFiltering.jl
+++ b/src/ImageFiltering.jl
@@ -32,9 +32,9 @@ export Kernel, KernelFactors,
     findlocalminima, findlocalmaxima,
     blob_LoG, BlobLoG
 
-FixedColorant{T<:Normed} = Colorant{T}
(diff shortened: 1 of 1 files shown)


timholy:
This is only cosmetic, as
```julia
julia> using FixedPointNumbers, ColorTypes

julia> FixedColorant{T<:Normed} = Colorant{T}
ColorantNormed (alias for Colorant{T} where T<:Normed)

julia> isconst(Main, :FixedColorant)
true
```
The key point is that all of these have curly braces `{T}` in the left-hand side, so Julia knows they refer to types and marks them `const`.

That said, if you like the code better with the explicit declarations, we can merge this.

kunzaatko:
Thanks for the explanation!
I do like the code better since I see it often in other packages and consider it a standard that supports explicitness.
However you should not feel pressured into merging this if you like the code the other way around.

